PR #22, opened 2015-02-05
Visual Studio 2013 and simple .NET support

LennardF1989:
Heya,

In my fork, I added out-of-the-box support for Visual Studio 2013, for those who just want to checkout, open up and compile.

I separated the project in librsync and rdiff, where librsync compiles to a Static Library, and rdiff links to that. By a small modification to the rdiff.c file. It can optionally expose the three rdiff functions we know from the executable (signature, delta, patch) to a DLL, which in return can be Pinvoked in .NET again (example provided). It's obviously also possible to output rdiff to a Static Library as well, which allows you to use it as an easy access wrapper for use in your C++ projects.

I'm leaving a pull request here to bring it to attention, maybe there is some useful stuff in there which you could pull back?

Lennard


sourcefrog:
Thanks!

You seem to have added a copy of all of the source under PCBuild? Could you avoid that?

I'd really rather not ship a copy of popt - can you build it separately?

LennardF1989:
@sourcefrog If you'd like I can make a proper integration by the use of flags and such. Since I wasn't sure this was something you'd like I kept the Visual Studio specific changes outside the main code, which in theory would make upgrading easier. However, if you give me a green flag for liking this, I'll make a proper pull request which works right of the main branch.

However, you I seperated rdiff from librsync and made a build-configuration around that in regards to outputting a DLL and such, I have no idea how I could do that for Linux (another reason as to why I kept a lot of the stuff in PCBuild, as it's Visual Studio specific).

As for popt, I will have to create a simple batch file that does the same the pyhton script and the like for the simple fat: Windows people generally don't like installing all kinds of extra tool just so the can run a silly script like that.

If you can give me instructions as to what you'd like to see I'll see what I can do to provide them.

sourcefrog:
Do you have an opinion about the switch to CMake proposed in https://github.com/librsync/librsync/pull/27? Would that be a good (or even decent) solution for Windows?

> As or popt: The reason I included this is because I simply cannot find a working piece of source, just the one I found in the dropbox fork.

Ouch, I guess it has really died since I used it here. Let's ship it then in the top level.

I like the idea of having library functions that do the same as rdiff, including on unix.

Salamek:
@sourcefrog @LennardF1989 cmake can generate VS project files: `cmake -G "Visual Studio 12"` that should be all,  but it is not tested for #27 
@ popt issue, maybe time to move to something else ? http://stackoverflow.com/questions/189972/argument-parsing-helpers-for-c-unix

sourcefrog:
> @ popt issue, maybe time to move to something else ?

Maybe it is. But, that question is all about Unix, and it seems like all the options there would have to be shipped in the tree anyhow. So it's not clear switching from popt would really gain anything. 

The question does actually link to the popt source distribution at http://rpm5.org/files/popt/

sourcefrog:
I'm going to close this: I appreciate you putting it up and I would like a smoothly working MSVC build, but this is probably largely obsolete with the CMake change. 

I don't  have an MSVC install handy to test this. If someone can, please let me know or put up a separate pr to fix it.

LennardF1989:
Is it possible to still include the rdiff.c change which separates the command-line and actual functions so they can be used externally?

https://github.com/LennardF1989/librsync/blob/master/PCbuild/rdiff.c

This way all rdiff gets multiple uses:
- As a library on top of librsync, for easy access.
- As a separate DLL, which in turns can for example be P/Invoke it in C#, which is very useful :)
- As a executable, turning it into a command-line tool.

Right now its limited to just the last due to the `poptContext opcon` argument, which is a shame for such a handy wrapper!

sourcefrog:
I recently merged another change that makes rdiff optional. It would be
good to also expose the high level functions from rdiff but not the main
routine, if that's what you mean.

LennardF1989:
I basically want these 3 methods exposed in rdiff:

    rs_result rdiff_patch(const char *baseFile, const char *deltaFile, const char *newFile)
    rs_result rdiff_delta(const char *sigFile, const char *newFile, const char *deltaFile)
    rs_result rdiff_sig(const char *baseFile, const char *sigFile)

So if that is what you mean, then I'm game :P

I had a whole reasoning typed out, but I think it's pretty clear: One doesn't always need the low-level functions and just 3 high-level ones that do what librsync is good at while handling all the nitty gritty work of loading files :) 

Besides, not only can you statically link librsync into your rdiff, it will also result in a much smaller dependency. Even more so if you get rid of popcontext functions and references when not build as a console application.

sourcefrog:
On Tue, Feb 9, 2016 at 12:43 PM, Lennard Fonteijn <notifications@github.com>
wrote:

> I basically want these 3 methods exposed in rdiff:
>
> rs_result rdiff_patch(const char *baseFile, const char *deltaFile, const char *newFile)
> rs_result rdiff_delta(const char *sigFile, const char *newFile, const char *deltaFile)
> rs_result rdiff_sig(const char *baseFile, const char *sigFile)
>
> So if that is what you mean, then I'm game :P
>
Yes that sounds good to me.

PR #24, opened 2015-02-21
Unknown rdiff options -i and -z

robert-scheck:
Patch by Daniel Baumann <daniel@debian.org> for librsync >= 0.9.7, which makes rdiff aware of -i and -z getopt options mentioned in --help output. For further information, please have a look to Debian bug ID #435894 at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=435894


sourcefrog:
I'm not sure it's worth keeping them here but unimplemented, but this is an improvement, thanks.

PR #27, opened 2015-05-11
Implementation of cmake build system and cleanup

Salamek:
Hi there!

Because i rly dont like autotools, and i was unable to build librsync with it even under Linux i decided to implement cmake build system + some cleanup of files:

* Moved *.c *.h files into src dir
* Fixed/updated doxygen
* Renamed testsuit to tests
* Bunch of ignores in .gitignore
* Deleted lot of ~useless files  (i hope)
* And various things all around (see diff)
* Updated README.md
* Tests are now run by `make test` ~using CTest

Tested under:

* Linux x86_64 (gcc, clang)
* MAC OS X 10.10.3 (gcc)
* Windows XP (mingw)

This 3 platforms are all i need and all i can test on...

I hope this changes will be useful somehow for you guys.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ html
 latex
 Testing
 *.user
+.sw?
+.*.sw?
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,93 +159,81 @@ if(DOXYGEN_FOUND)
 endif(DOXYGEN_FOUND)
 
 # Generate prototab.c/h
-execute_process(COMMAND ${PERL_EXECUTABLE} "${CMAKE_SOURCE_DIR}/src/mkprototab.pl" "${CMAKE_SOURCE_DIR}/src/prototab.c" "${CMAKE_SOURCE_DIR}/src/prototab.h")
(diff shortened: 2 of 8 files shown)


sourcefrog:
Hi, thanks for the patch.

I'm open to switching to a different build system, automake is not my favorite thing either. 

This one mega patch is going to be kind of a pain to review but perhaps I can pick them out from your branch.

Salamek:
@sourcefrog yea, sorry for this "mega patch" i had to make it working and fit it into our codestyle ASAP, just pick what do you need and like and implement it... but i think that ~every change i made is ~reasonable well some may need some "polishing" :-)

sourcefrog:
I'm trying to decide if cmake is the way to go vs scons (which I used before but know has limitations), waf, maybe something else? I would like something with less hairy implementation than automake and that will work on Windows and Unix. I guess cmake fits the bill.

paulharris:
+1 for cmake, its become the standard from my POV and it works well with MSVC and great on linux with "ninja" (much better than "make").

sourcefrog:
merged! :checkered_flag:

Salamek:
Coool && THX!

PR #29, opened 2015-06-29
Added test to show weird behaviour with small output buffers.

paulharris:
Backported from my CMake-based branch.

I see issues when the output buffer is too small.
librsync.h indicates I should increase the buffer size and reexecute, but that seems to be buggy.

Here is a test, the parameters control the input size, the initial output buffer size, and the output buffer growth rate (additive).

Its quite easy to allocate too-small a buffer, and as the input size goes up, it seems that the minimum required output buffer sizes also change.

### DIFF
--- a/tests/test_small_output.c
+++ b/tests/test_small_output.c
@@ -1,0 +1,68 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <librsync.h>
+
+// Test behaviour when the output buffer is too small, and grows bit by bit
+// Parameters: test_small_output.test  INPUTSIZE  INITIAL_BUFFER_SIZE  OUTPUT_GROWTH_RATE
+
+int main(int argc, const char** argv)
+{
+   int i;
+   size_t input_size = atoi(argv[1]);
+   size_t start = atoi(argv[2]);
+   size_t step = atoi(argv[3]);
+
+   printf("Input size: %d\n", input_size);
+   printf("Output buffer start at %d step growth of %d\n", start, step);
+
+   // make this silly-huge, we don't want to test part
+   char * outbuf = (char*)malloc(1024+10*input_size);
+
+   // load with anything
+   char * in = (char*)malloc(input_size);
+   for (i = 0; i < input_size; ++i)
+      in[i] = 'x';
+
+   // generate a signature
+   rs_buffers_t buf;
+   buf.next_out = outbuf;
+   buf.avail_out = start;
+   buf.next_in = (char*)(in);   // TODO why is this not const void* ?
+   buf.avail_in = input_size;
+   buf.eof_in = 0;
+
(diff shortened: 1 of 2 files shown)


sourcefrog:
Thanks.

Is there already a bug for this?

Are you saying this fails for some parameter values?

paulharris:
Hi,

uuh I don't think there is a bug for this,
and yes it fails for many parameters.

Smallest test case fails,
where the input is zero bytes,
initial output buffer is 1 byte,
and output buffer is grown by 1 byte at a time.

Biggest test case that fails is where the
input is 2,000,000 bytes
initial output buffer is 16000 bytes
output buffer is grown by 30 bytes at a time.

These numbers are just picked out of the air by me,
the input buffer is just filled with 'x' characters,

This patch has a script you can use to compile and run, you can see the parameters in the shell script there.   I also have a patch on top of my cmake system, but for the base rsync, the procedure is:

git checkout -b issue-small-output paulharris/issue-small-output
./autogen.sh && ./configure && make && (cd tests && ./test_small_output.run)
and to repeat,
cd tests
./test_small_output.run

To run the small testcase, for example, its:
./test_small_output.test 0 1 1
big test case:
./test_small_output.test 2000000 16000 30

output for small:
$ ./test_small_output.test 0 1 1
Input size: 0
Output buffer start at 1 step growth of 1
FAIL, expected 'done'

Show success (just a bigger initial buffer):
$ ./test_small_output.test 0 12 1
Input size: 0
Output buffer start at 12 step growth of 1
Final signature size: 12

sourcefrog:
Thanks.

Is there already a bug for this?

Are you saying this fails for some parameter values?

paulharris:
strange, I see the same comment of yours twice.

btw I am not sure how I go about making a bug and attaching this pull request...

sourcefrog:
So it looks like this has the test driver in the tree, but not run by cmake (probably because this was backported), and it feels like tests that aren't hooked in will just rot.  I'm glad you wrote this but it might be better to leave it attached to the bug until the bug itself is fixed, rather than merging in either a failing or not-running state.

tests/test_small_output.c
```diff
@@ -0,0 +1,68 @@
...
+   // load with anything
+   char * in = (char*)malloc(input_size);
+   for (i = 0; i < input_size; ++i)
+      in[i] = 'x';
+
+   // generate a signature
+   rs_buffers_t buf;
+   buf.next_out = outbuf;
+   buf.avail_out = start;
+   buf.next_in = (char*)(in);   // TODO why is this not const void* ?
```
sourcefrog: Moving all the data pointers to `void*` would be a worthwhile separate change, but probably counts as an API break.

tests/test_small_output.run
```diff
@@ -0,0 +1,39 @@
+#! /bin/sh
```
sourcefrog: I'd like to avoid adding new sh tests and go towards something that'll run more easily on Windows (#82).  Specifically here I would just put them inside the C test file and run that directly.

sourcefrog:
I'm also going to close the PR and leave the bug open until we have a fix.  Thanks again.

PR #31, opened 2015-06-30
Add comments and asserts regarding rs_build_hash_table()

paulharris:
It is not clear in the header file that rs_build_hash_table() MUST be
called after the signature has been loaded.


paulharris:
This is for fixing #31

paulharris:
Rebased, is that ok?

sourcefrog:
Looks good, thanks!

Issue #33, opened 2015-06-30
rs_job_iter needs to be called twice with eof=1


paulharris:
This smells like a bug,
please see pull request with demonstration and comments.

sourcefrog:
It does smell suspicious; thanks for the test case.

dbaarda:
FTR I checked this against master and my recent pull request #119 and it's still there.

Arguably it's not a bug; the documented API says you need to keep calling rs_job_iter() until it returns RS_DONE (or an error). However, it is something that could be done better.

sourcefrog:
thanks @dbaarda !

PR #34, opened 2015-06-30
Demonstrate issue with generating delta (extra call required)

paulharris:
Need to call rs_job_iter with EOF=1 twice,
which seems strange as the internal code has a while loop and plenty of
checks, so I would have assumed 1 pass would have been enough.

Please see issue #33


sourcefrog:
Thanks for this too: I appreciate the test case but I think this doesn't make sense to merge by itself until it's hooked in to Cmake and probably until the bug is fixed.  It's useful to have this for context on the bug.

Issue #35, opened 2015-07-01
Crash with badly behaving rs_copy_cb callback


paulharris:
I wrote a test which can make librsync get confused and request out-of-bounds data,
and write past the end of the output buffer.

    test_delta_read.test: test_delta_read.c:34: read_memory_callback: Assertion `pos < mem->size' failed.
Aborted

```c
// in:
static rs_result rs_patch_s_copying(rs_job_t *job) {
...;
// the code uses the callback's returned 'len' value without checking if its larger than the output buffer
// ie can write past the end of the output buffer.
    result = (job->copy_cb)(job->copy_arg, job->basis_pos, &len, &ptr);
// nothing checked with len, just memcpy blindly...
    memcpy(buffs->next_out, ptr, len);
```

2 commits coming in a pull request.  1 with the test, 1 with my fix.

PR #36, opened 2015-07-01
Handle copy cb returning more data than requested

paulharris:
Addresses #35

### DIFF
--- a/patch.c
+++ b/patch.c
@@ -209,13 +209,14 @@ static rs_result rs_patch_s_copy(rs_job_t *job)
 static rs_result rs_patch_s_copying(rs_job_t *job)
 {
     rs_result       result;
-    size_t          len;
+    size_t          len, requested_len;
     void            *buf, *ptr;
     rs_buffers_t    *buffs = job->stream;
 
     /* copy only as much as will fit in the output buffer, so that we
      * don't have to block or store the input. */
     len = (buffs->avail_out < job->basis_len) ? buffs->avail_out : job->basis_len;
+    requested_len = len; /* remember for asserting correct behaviour */
 
     if (!len)
         return RS_BLOCKED;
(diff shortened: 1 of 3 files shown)


sourcefrog:
We need to hook the test into `configure.ac` so it actually runs.

sourcefrog:
It looks like I should also clarify in the docs that the length passed in is a maximum.

sourcefrog:
I wonder if we should just use only the part we want, even if the caller returned too much.  It could be just a warning.

sourcefrog:
I merged the equivalent of the fix in https://github.com/librsync/librsync/commit/088600fb8f51cdacbbd8143b5fffc17994e838bf but we should still merge the test.

sourcefrog:
I have an updated version of this test in https://github.com/sourcefrog/librsync/tree/copy-cb2 but it's failing, despite the original bug apparently being fixed.

dbaarda:
I just had a look at this in the interests of closing out this pull request.

IMHO a badly behaving callback is a programming error that violates the contract of the API. As such, we should not be adding defensive code that will be an overhead for good code in release builds. However, we should be adding asserts() that check the API contract is being honoured for debug builds.

I also spotted some inefficiency in that piece of code that should be fixed... So I might just fix that now.

dbaarda:
Actually, I just looked at the current code (instead of just this patch), and I see the inefficiency has been fixed (unnecessary alloc/memcpy) and the code now just uses only the desired len if the callback returned more.

I still think asserts would have been better, but changing it now would arguably be a regression, so my motivation for doing anything with this has faded.

As I consider correctly handling a bad callback a bit of a non-feature, I don't really care about adding tests to check that it works... so IMHO this pull request should be dropped.

paulharris:
ok

dbaarda:
Looking a little closer I spotted another little bug which would make patch hang forever for an invalid delta file with a zero length copy instruction.

Also, the current code only emits a trace "warning message" when this happens, which means for builds with trace turned off it becomes invisible. Even with trace turned on it's likely to be missed in the trace noise. As this is likely an error developers should know about, I think it needs to be more visible than that.

So I've put together a patch that;

1. fixes the little buglet for bad patch files.
2. Uses assert() to check the copied length is valid so debug builds fail with an assert error.
3. Keeps the backwards-compatible defensive handling for non-debug builds but logs a warning.
4. Tidies up the trace output a little bit to only output the callback return value for errors.

At some stage in the future we might be able to remove 3. but I thought it wisest to leave it in for now. See #206 for details.

dbaarda:
I'm pretty happy with #206 and don't think we need tests to exercise assert() API contract checking. 

I'm going to abandon this pull request.

Issue #50, opened 2015-12-01
Crash in `rs_search_for_block`


mmasenlle:
Sometimes the l and r variables take the value sig->count when leaving de while loop. Then this lines are executed:

```
if (l == r) {
int i = sig->targets[l].i;
rs_block_sig_t *b = &(sig->block_sigs[i]);
```

Tested with compilers 
gcc (i686-posix-dwarf-rev1, Built by MinGW-W64 project) 4.9.2
gcc (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 5.2.0
the i var goes out of bounds

In Linux with differents gcc versions i get a value that is bound.

jellysheep:
The bug is present on Linux as well. On an ARMv7 with GCC 5.2.0, I get the following error (when using librsync via Python/rdiff-backup):
```
Program received signal SIGSEGV, Segmentation fault.
rs_search_for_block (weak_sum=1858441517, inbuf=0x2 <error: Cannot access memory at address 0x2>, block_len=block_len@entry=3232, sig=0x54d1f220, stats=stats@entry=0x54d1f8f0, 
    match_where=0x7efee080, match_where@entry=0x7efee078) at /home/ladmin/librsync/src/librsync-2.0.0/src/search.c:224
224             if (weak_sum != b->weak_sum)
(gdb) bt
#0  rs_search_for_block (weak_sum=1858441517, inbuf=0x2 <error: Cannot access memory at address 0x2>, block_len=block_len@entry=3232, sig=0x54d1f220, stats=stats@entry=0x54d1f8f0, 
    match_where=0x7efee080, match_where@entry=0x7efee078) at /home/ladmin/librsync/src/librsync-2.0.0/src/search.c:224
#1  0x761c1bf4 in rs_findmatch (match_len=<synthetic pointer>, match_pos=0x7efee078, job=0x54d1f840) at /home/ladmin/librsync/src/librsync-2.0.0/src/delta.c:273
#2  rs_delta_s_scan (job=0x54d1f840) at /home/ladmin/librsync/src/librsync-2.0.0/src/delta.c:157
#3  0x761c2abc in rs_job_work (job=0x54d1f840, buffers=buffers@entry=0x7efee118) at /home/ladmin/librsync/src/librsync-2.0.0/src/job.c:187
#4  0x761c2b38 in rs_job_iter (job=<optimized out>, buffers=0x7efee118) at /home/ladmin/librsync/src/librsync-2.0.0/src/job.c:145
[...]

PR #54, opened 2016-01-10
Run Travis tests on several interesting environments

sourcefrog:


sourcefrog:
This makes the 'mutate' test run on OS X, which is intermittently failing due to #50. The bug doesn't seem to manifest on Linux on Travis.

PR #59, opened 2016-01-19
Handle searched checksum is bigger than any existing

VictorDenisov:
Fix for issue #50

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -221,7 +221,7 @@ rs_search_for_block(rs_weak_sum_t weak_sum,
             r = m;
     }
 
-    if (l == r) {
+    if ((l == r) && (l <= bucket->r)) {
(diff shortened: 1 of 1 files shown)


src/search.c
```diff
@@ -221,7 +221,7 @@ rs_search_for_block(rs_weak_sum_t weak_sum,
             r = m;
     }
 
-    if (l == r) {
+    if ((l == r) && (l <= bucket->r)) {
```
sourcefrog: As you say in the summary, this is the case where the weak sum is greater than any that exist, so we walk off the right end of the range - I guess by line 219 `l = m + 1`. (Though, since `r` starts out at `bucket->r + 1` perhaps we can hit it that way, but I think not.)

Would it be simpler to just `return 0` at the point where we would update `l`?

Issue #8, opened 2014-07-31
Please add gzip/bzip compression for delta files in rdiff


pavel-odintsov:
Hello!

I tried to use flags --gzip/--bzip for rdiff but got error:

``` bash
rdiff: ERROR: (rdiff_options) sorry, compression is not really implemented yet
```

For my data (VPS disks) compression provides really excellent compression for delta files:

``` bash
source size: 4.6 Gb delta size: 2093.0 MB compressed size: 223.0
source size: 14.8 Gb delta size: 2205.0 MB compresses size: 998.7 MB
```

Thank you!

dbaarda:
I'm considering tackling this next. Either that or Rabin-Karp rollsums... whichever people prefer.

Note that signature files being collections of hash values probably don't compress at all well, unless they have long runs of identical blocks. I'm planning to only add compression to the deltas, with optional "context compression" support (which compresses hits as well as misses to prime the compressor with context from matching blocks).

yxj1992:
I have set  `cmake -D ENABLE_COMPRESSION=ON .`,
but it doesn't work，'ERROR: (rdiff_options) sorry, compression is not really implemented yet'，
who can tell me why?

dbaarda:
yxj1992: because that feature hasn't been implemented yet.

dbaarda:
FTR, I found a good comparison between gzip, bz2, and xz here;

https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/

Looking at this xz is the clear winner on compression ratio, but it is still diminishing returns against gzip compared to no compression at all. The clear winner on speed AKA cpu is gzip, particularly for decompression. For bz2, it beats gzip on compression, but it's not as good as xz, and pays a nasty speed/cpu price.

For librsync's application, I feel gzip is the winner, and there's not much point in implementing support for xz or particularly bzip2.

sourcefrog:
For new code now it's probably worth looking at zstd too.

>

dbaarda:
I looked at zstd. I agree it looks like the best/only solution needed for compression. There is a Debian libzstd-dev package we can just build/link against.

On how to best implement this, I think this is blocked on implementing a hit/miss callback api for deltas as described in #197. That API would make it much easier to implement different kinds of delta output formats including things like compression and whole-file-checksums, as different callback implementations.

dbaarda:
[...]

PR #60, opened 2016-01-20
Build rdiff target optionally

mbrt:
I've tweaked the build process a bit. I have the need to minimize build dependencies, and so I made the following changes:
* Added an option (default ON) to build the `rdiff` target;
* Added an option (default OFF) to build with compression support;
* Made `zlib` and `bzip2` libraries optional.

This allows me to build only the `rsync` library (and not `rdiff`), and thus avoid having `popt` dependency. Moreover, `zlib` and `bzip2` use can be disabled.

The removal of `REQUIRED` attribute in `find_package` turns an hard error in a soft one, and does not stop the generation step. This allows the user to tweak parameters after generation with CMake.

The compression support is not implemented for now, so I've kept it OFF by default. When the support is present, you can make the option ON by default!

Let me know what you think.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,11 @@ if (NOT CMAKE_SYSTEM_NAME)
 	message(FATAL_ERROR "No target OS set")
 endif()
 
+# Add an option to exclude rdiff executable from build
+# This is useful, because it allows to remove POPT dependency if a user is interested only in the
+# rsync library itself and not in the rdiff executable
+option(BUILD_RDIFF "Whether to build rdiff executable or not" ON)
+
 include ( CheckIncludeFiles )
 check_include_files ( alloca.h HAVE_ALLOCA_H )
 check_include_files ( dlfcn.h HAVE_DLFCN_H )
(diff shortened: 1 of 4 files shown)


mbrt:
Mmm... Not sure about the failure under clang-apple. I don't think it's my fault. Master seems broken too.

sourcefrog:
Sounds good to me.

> Mmm... Not sure about the failure under clang-apple in travis. I don't think it's my fault. Master seems broken too.

Yes, that's #50, not your fault.

CMakeLists.txt
```diff
@@ -37,6 +37,14 @@ if (NOT CMAKE_SYSTEM_NAME)
 	message(FATAL_ERROR "No target OS set")
 endif()
 
+# Add an option to exclude rdiff executable from build
+# This is useful, because it allows to remove POPT dependency if a user is interested only in the
+# rsync library itself and not in the rdiff executable
+option(BUILD_RDIFF "Whether or not to build rdiff executable" ON)
```
sourcefrog: Please document this in `install.md`

CMakeLists.txt
```diff
@@ -37,6 +37,14 @@ if (NOT CMAKE_SYSTEM_NAME)
...
+
+# Add an option to include compression support
```
sourcefrog: Please mention https://github.com/librsync/librsync/issues/8 here, that it's not actually functional

CMakeLists.txt
```diff
@@ -37,6 +37,14 @@ if (NOT CMAKE_SYSTEM_NAME)
 	message(FATAL_ERROR "No target OS set")
 endif()
 
+# Add an option to exclude rdiff executable from build
+# This is useful, because it allows to remove POPT dependency if a user is interested only in the
+# rsync library itself and not in the rdiff executable
+option(BUILD_RDIFF "Whether or not to build rdiff executable" ON)
+
```
sourcefrog: Please try adding in .travis.yml a config that builds without `rdiff`.

CMakeLists.txt
```diff
@@ -262,15 +287,20 @@ install(TARGETS rsync ${INSTALL_TARGETS_DEFAULT_ARGS} DESTINATION lib)
...
+  else (POPT_FOUND)
+    message (WARNING "Popt library is required for rdiff target")
+  endif (POPT_FOUND)
 
-install(TARGETS rdiff ${INSTALL_TARGETS_DEFAULT_ARGS} DESTINATION bin)
+  install(TARGETS rdiff ${INSTALL_TARGETS_DEFAULT_ARGS} DESTINATION bin)
+endif (BUILD_RDIFF)
 
 
 ########### install files ###############
```
sourcefrog: Please mention this and credit yourself in NEWS.md

sourcefrog:
That looks like a really worthwhile change, thanks. I'd just like a few tweaks.

mbrt:
I've addressed your comments. There is only one additional change. I've disabled some tests when `BUILD_RDIFF` is disabled, because they depend on `rdiff` executable! This is unfortunate, but the alternative would be to have those tests to fail... Let me know!

sourcefrog:
Yeah, unfortunately past-me wrote most of the tests on top of rdiff rather than as C-level unit tests. Probably better to disable them for now though it does make this a bit more risky. Maybe we should note that in install.md?

mbrt:
I've disabled them in e215215. I think the risk is not very much, since the default build is with `rdiff`, and we have also travis which tests all the configurations. However it's better to add a note in install.md, as you said!

mbrt:
Added the note in the last commit.

sourcefrog:
Great, thanks.

PR #61, opened 2016-01-25
Fix #50: don't bisect off the end of the bucket

### LINKED ISSUE #50: Crash in `rs_search_for_block`

sourcefrog:


sourcefrog:
@VictorDenisov please have a look at this - I think it is equivalent but to me easier to understand

sourcefrog:
discarded in favor of #59

PR #64, opened 2016-01-31
Fixed MSVC build under Windows

mbrt:
Removed `librsync.h` include from `version.c`. The build was failing
because `stdlib.h` include was missing, and so `size_t` type not found.
There were two options to fix that: either remove `librsync.h` include,
or add also `stdlib.h` include. Since they were both not used by
`version.c`, I've chosen to remove them.

### DIFF
--- a/src/version.c
+++ b/src/version.c
@@ -20,7 +20,6 @@
  */
 
 #include "config.h"
-#include "librsync.h"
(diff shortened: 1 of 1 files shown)


sourcefrog:
thanks

PR #67, opened 2016-03-29
Various improvements

gulikoza:
Various improvements I have in my branch.
I have tested and built librsync in MinGW. The realloc calling for every signature block is extremely slow...either this is MinGW specific or nobody noticed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@
 project(librsync)
 cmake_minimum_required(VERSION 2.6)
 
+INCLUDE(CMakeDependentOption)
 
 set(LIBRSYNC_MAJOR_VERSION 2)
 set(LIBRSYNC_MINOR_VERSION 0)
@@ -36,11 +37,6 @@ endif()
 if (NOT CMAKE_SYSTEM_NAME)
 	message(FATAL_ERROR "No target OS set")
 endif()
-
-# Add an option to exclude rdiff executable from build
-# This is useful, because it allows to remove POPT dependency if a user is interested only in the
-# rsync library itself and not in the rdiff executable
-option(BUILD_RDIFF "Whether or not to build rdiff executable" ON)
 
 # Add an option to include compression support
 option(ENABLE_COMPRESSION "Whether or not to build with compression support" OFF)
@@ -151,6 +147,11 @@ find_package(POPT)
 if (POPT_FOUND)
   include_directories(${POPT_INCLUDE_DIRS})
 endif (POPT_FOUND)
+
+# Add an option to exclude rdiff executable from build
+# This is useful, because it allows to remove POPT dependency if a user is interested only in the
+# rsync library itself and not in the rdiff executable
+cmake_dependent_option(BUILD_RDIFF "Whether or not to build rdiff executable" ON "POPT_FOUND" OFF)
 
 # Find BZIP
 find_package (BZip2)


gulikoza:
Refreshed pull to include test fixes which now pass successfully.

sourcefrog:
Thanks for that.

In future I'd prefer smaller PRs with one change each.

src/fileutil.c
```diff
@@ -63,6 +63,15 @@ rs_file_open(char const *filename, char const *mode)
 	    return stdin;
     }
 
+    if (!force && is_write) {
+	if ((f = fopen(filename, "rb"))) {
```
sourcefrog: Rather than look-before-you-leap, it'd be better to open the file in a way that won't overwrite existing files, probably using `open` and then `fdopen`.

tests/check-rdiff
```diff
@@ -34,9 +34,9 @@ sub run {
 srand(0);
 makefile('old',$size);
 makefile('new',$size);
-run("time rdiff signature -b $blocklen old old.sig");
-run("time rdiff delta -s old.sig new delta");
-run("time rdiff patch -s old delta new2");
+run("time rdiff signature -f -b $blocklen old old.sig");
```
sourcefrog: Can you please test the mode without `-f`?

src/rdiff.c
```diff
@@ -100,6 +100,7 @@ const struct poptOption opts[] = {
     { "stats",        0,  POPT_ARG_NONE, &show_stats },
     { "gzip",        'z', POPT_ARG_NONE, 0,             OPT_GZIP },
     { "bzip2",       'i', POPT_ARG_NONE, 0,             OPT_BZIP2 },
+    { "force",       'f', POPT_ARG_NONE, &file_force },
```
sourcefrog: Thanks, that sounds good.  Please mention it in `NEWS.md`

src/buf.c
```diff
@@ -93,6 +95,15 @@ void rs_filebuf_free(rs_filebuf_t *fb)
 }
 
 
+void rs_get_filesize(FILE *f, rs_long_t *size)
+{
+    rs_long_t pos = ftell(f);
+    fseek(f, 0, SEEK_END);
```
sourcefrog: This works but it seems ugly to seek just to get the size. I'd use `fstat(fileno(...)...)`

src/buf.c
```diff
@@ -93,6 +95,15 @@ void rs_filebuf_free(rs_filebuf_t *fb)
 }
 
 
+void rs_get_filesize(FILE *f, rs_long_t *size)
```
sourcefrog: This would be better in `util.c`

src/buf.c
sourcefrog: Also, this needs to allow for reading from a pipe where we can't get the length.

src/readsums.c
```diff
@@ -56,18 +56,25 @@ static rs_result rs_loadsig_add_sum(rs_job_t *job, rs_strong_sum_t *strong)
     rs_signature_t      *sig = job->signature;
     rs_block_sig_t      *asignature;
 
-    sig->count++;
-    new_size = sig->count * sizeof(rs_block_sig_t);
+    if (sig->count <= job->stats.sig_blocks) {
```
sourcefrog: It seems strange to look at `stats` here which is meant to be just a side effect.  What are you trying to check?

src/whole.c
```diff
@@ -117,6 +117,10 @@ rs_loadsig_file(FILE *sig_file, rs_signature_t **sumset, rs_stats_t *stats)
     rs_result           r;
 
     job = rs_loadsig_begin(sumset);
+
+    /* Estimate a number of signatures by file size */
+    rs_get_filesize(sig_file, &job->sig_size);
```
sourcefrog: The field ought to be initialized eg to 0 in case the length can't be measured.

src/buf.c
gulikoza: This is more portable, fstat would need another set of different functions depending on the platform. We already have ftell and fseek supposedly supporting large files. Sure, it's unlikely signature would be larger than 4GB, but it doesn't hurt to be on the safe side.

The algorithm falls back to realloc if size is not available, so it should work for pipes...

src/buf.c
gulikoza: fseek and ftell are defined in buf.c, I didn't want to move (or duplicate) those defines...

src/job.h
```diff
@@ -49,6 +49,7 @@ struct rs_job {
     /** Signature that's either being read in, or used for
      * generating a delta. */
     rs_signature_t      *signature;
+    rs_long_t           sig_size;
```
sourcefrog: Please add a comment.

src/readsums.c
```diff
@@ -56,18 +56,25 @@ static rs_result rs_loadsig_add_sum(rs_job_t *job, rs_strong_sum_t *strong)
...
+
+        sig->count++;
```
sourcefrog: This seems to sometimes not increment `sig->count` at all...?

src/readsums.c
```diff
@@ -145,6 +152,14 @@ static rs_result rs_loadsig_s_stronglen(rs_job_t *job)
 
     job->statefn = rs_loadsig_s_weak;
     
+    /* Estimate the number of blocks stored in signature */
+    /* Magic+header is 12 bytes,
+       each block thereafter is 4 bytes weak_sum+strong_sum_len bytes */
+    if (job->sig_size > 0) {
+        job->signature->count = (job->sig_size-12)/(4+job->strong_sum_len);
```
sourcefrog: I think rather than putting this in `signature->count` I would have a separate `job->estimated_signature_count`, and keep the signature object for just what was actually read.

src/readsums.c
```diff
@@ -145,6 +152,14 @@ static rs_result rs_loadsig_s_stronglen(rs_job_t *job)
...
+        rs_trace("estimating %d blocks in signature", job->signature->count);
```
sourcefrog: Could you just allocate everything here if you now know the size?

sourcefrog:
Thanks, that's really worthwhile. I merged some and added comments on the others.

src/readsums.c
gulikoza: stats.sig_blocks is incremented for each signature block read.
sig->count is the estimated count of signatures (if set before based on the signature file size) or current count (if signature file size is not known). When sig->count <= job->stats.sig_blocks, we're out of memory (since at most sig->count blocks were allocated and we're trying to add a new block) so we need to reallocate, either because we miscalculated signature count or file size was not known in advance. This path is basically current path of librsync as if sig->count was never set in advance...

src/whole.c
gulikoza: rs_alloc_struct already zeroes out entire job structure when allocating...

src/readsums.c
gulikoza: Yep...because sig->count is set before in rs_loadsig_s_stronglen() to a total count when estimating the number of signatures. If signature size is not known before, this codepath increments it after adding every block of sums from the signature just like now.

src/readsums.c
gulikoza: I'll rework the algorithm to use job->estimated_signature_count as noted bellow. This will probably make it clearer. There's no point in having sig->count larger than current number of read blocks, I agree this should be a separate variable.

PR #70, opened 2016-05-01
Do not overwrite files by default.

gulikoza:
Add option --force (-f) to restore previous functionality and overwrite all files.

I closed the previous PR and split the remaining code into separate requests.
I'll try to address the comments from previous PR here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,11 @@ NOT RELEASED YET
 
  * `rdiff -s` option now shows bytes read/written and speed. (gulikoza)
 
+ * Running rdiff should not overwrite existing files (signatures, deltas and
+   new patched files) by default. If the destination file exists, rdiff will
+   now exit with an error. (gulikoza)
+   Add new option -f (--force) to overwrite existing files.
+
 ## librsync 2.0.0
 
 Released 2015-11-29
--- a/src/fileutil.c
+++ b/src/fileutil.c
@@ -49,7 +49,7 @@
  * \param fopen-style mode string.
  */
 FILE *
-rs_file_open(char const *filename, char const *mode)
+rs_file_open(char const *filename, char const *mode, int force)
 {
     FILE           *f;
     int		    is_write;
@@ -61,6 +61,15 @@ rs_file_open(char const *filename, char const *mode)
 	    return stdout;
 	else
 	    return stdin;
+    }
+
+    if (!force && is_write) {
+	if ((f = fopen(filename, "rb"))) {
+	    // File exists
+	    rs_error("File exists \"%s\", aborting!", filename);
+	    fclose(f);
+	    exit(RS_IO_ERROR);
+	}
     }
 
     if (!(f = fopen(filename, mode))) {
--- a/tests/delta.test
+++ b/tests/delta.test
@@ -33,9 +33,29 @@ for delta in $inputdir/*.delta
 do
     for inbuf in $bufsizes
     do
-        cmd="$bindir/rdiff -I$inbuf patch /dev/null $delta $new"
+        cmd="$bindir/rdiff -I$inbuf -f patch /dev/null $delta $new"
         run_test $cmd
         expect=$inputdir/`basename $delta .delta`.expect
 	check_compare $expect $new
     done
 done
+
+# Make sure rdiff can't overwrite files if ran without -f
+# RS_IO_ERROR == 100 should be returned if file exists
+RS_IO_ERROR=100
+output_expect="File exists \".*\", aborting!"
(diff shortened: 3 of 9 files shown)


src/fileutil.c
```diff
@@ -63,6 +63,15 @@ rs_file_open(char const *filename, char const *mode)
 	    return stdin;
     }
 
+    if (!force && is_write) {
+	if ((f = fopen(filename, "rb"))) {
```
gulikoza: "Rather than look-before-you-leap, it'd be better to open the file in a way that won't overwrite existing files, probably using open and then fdopen."

Using fdopen to switch to write mode 'do not cause truncation of the file.' FOPEN(3).
Closing and reopening the file in write mode seems the simplest solution ?

tests/delta.test
```diff
@@ -33,9 +33,15 @@ for delta in $inputdir/*.delta
...
         expect=$inputdir/`basename $delta .delta`.expect
 	check_compare $expect $new
     done
 done
+
+# Make sure rdiff can't overwrite files if ran without -f
+# RS_IO_ERROR == 100 should be returned if file exists
+cmd="$bindir/rdiff patch /dev/null $inputdir/01.delta $new"
+echo "    $cmd" >&2
+$cmd 2>&1 | grep -q 'File exists' || fail_test "$?" "File exists not returned when overwriting files without --force"
```
gulikoza: Added test that should print proper error when file exists and no --force was given.

src/fileutil.c
sourcefrog: Personally I would just use open/fdopen unless you think that's going to be annoying on Windows?

tests/delta.test
```diff
@@ -33,9 +33,15 @@ for delta in $inputdir/*.delta
...
-        cmd="$bindir/rdiff -I$inbuf patch /dev/null $delta $new"
+        cmd="$bindir/rdiff -I$inbuf -f patch /dev/null $delta $new"
         run_test $cmd
         expect=$inputdir/`basename $delta .delta`.expect
 	check_compare $expect $new
     done
 done
+
+# Make sure rdiff can't overwrite files if ran without -f
+# RS_IO_ERROR == 100 should be returned if file exists
```
sourcefrog: Can you please check that we get that return code, and that the file is not changed?

NEWS.md
```diff
@@ -36,6 +36,10 @@ NOT RELEASED YET
 
  * `rdiff -s` option now shows bytes read/written and speed. (gulikoza)
 
+ * Rdiff should not overwrite files (signatures, deltas and patched files)
```
sourcefrog: The style is just lowercase `rdiff`.

src/fileutil.c
```diff
@@ -63,6 +63,15 @@ rs_file_open(char const *filename, char const *mode)
...
+	    // File exists
+	    rs_error("File exists \"%s\", aborting!", filename);
+	    fclose(f);
+	    exit(RS_IO_ERROR);
```
sourcefrog: Wow, that's annoying I didn't make this return a status.  Still, probably better to return NULL than to exit the whole program.

gulikoza: Failure to open the file for processing a few lines down does the same thing: exit(RS_IO_ERROR);

As far as I can check, there are no checks if rs_file_open() has failed and returned NULL, returning it and continuing execution will probably segfault later because of NULL file pointer.

Exiting and returning RS_IO_ERROR seems consistent with the current code.

src/fileutil.c
gulikoza: fdopen() is POSIX not ANSI so it hurts portability.
ftruncate() is again problematic, it fails silently on Windows if file size >2 GB.

Closing and reopening really seems like the simplest solution :-)

gulikoza:
Refreshed pull request with an updated test that checks return code, expected program output and checks that the file has not changed.

gulikoza:
I have refreshed the pull request and improved the test a bit more.
Now $new content is changed, the last rdiff patch command is rerun and comparison is ran against the changed content. This tests that $new content really isn't changed by rdiff in any way, not just obvious errors like truncated output...

BTW, AppVeyor test fails with "Specify which project or solution file to use because the folder contains more than one project or solution file." Can you check that?

src/fileutil.c
gulikoza: I can use stat to check if the file exists?

sourcefrog:
Thanks.

> BTW, AppVeyor test fails with "Specify which project or solution file to use because the folder contains more than one project or solution file." Can you check that?

Yes, my configuration there is not right yet. It's not related to this patch.

Issue #71, opened 2016-05-02
triple.test is broken


gulikoza:
There is no $all_hashopts, so none of the tests actually run.

gulikoza:
Fixed in 0b7f876.

PR #72, opened 2016-05-02
Open stdout and stdin in _O_BINARY mode on Windows.

gulikoza:
As part of realloc patch, I wrote a test for rdiff using pipes. I discovered a CRITICAL problem! On Windows ALL output written to stdout has LF changed to CRLF. This CORRUPTS binary data and all signatures and deltas written to stdout are wrong!

Since this test was not running, I also added a quick fix for issue #71


sourcefrog:
Thanks.

Maybe we can set up Appveyor or some other Windows CI.

sourcefrog:
Let's see if https://ci.appveyor.com/project/sourcefrog/librsync works.  It may need some configuration.

PR #76, opened 2016-06-07
Small crosscompiling issues

sgh:
Hi

Please merge my changes - or maybe comment on any issues I have missed noticing.

Thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,10 +163,12 @@ if (BZIP2_FOUND)
 endif (BZIP2_FOUND)
 
 # Find Perl
-find_package (Perl REQUIRED)
-if (PERL_FOUND)
-  message (STATUS "PERL_EXECUTABLE  = ${PERL_EXECUTABLE}")
-endif (PERL_FOUND)
+if (NOT CMAKE_CROSSCOMPILING)
+	find_package (Perl REQUIRED)
+	if (PERL_FOUND)
+	  message (STATUS "PERL_EXECUTABLE  = ${PERL_EXECUTABLE}")
+	endif (PERL_FOUND)
+endif()
 
 # Find ZLIB
 find_package (ZLIB)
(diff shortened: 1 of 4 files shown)


sourcefrog:
Don't worry about the Appveyor build, it's not working yet.

Rather than renaming `buf.h`, I'd rather just move the contents into `librsync.h` and have that as the only public header.  I'd rather if you did one issue per PR.  If it's going to be public it should have comments, which might be able to just move from the `.c` file.

sgh:
Thanks for the comments. I do not think that the move of the declaration is necessary after all. My usecase involved storing and loading deltas, signatures and patches from in-memory buffers. I recently found out that fmemopen can solve that problem for me.

Do you agree with that - or do you still think that the buf.h declarations should be moved to librsync.h? In case it should be moved I'll happily fix the things you mention and issue new pull requests.

sourcefrog:
You should be able to do that using the existing buffer management in
librsync.h.

sourcefrog:
Please put up a new pr skipping detection of Perl when we don't intend to run the tests on the host machine. 

Let's discuss separately in an issue how you should handle memory buffers.

CMakeLists.txt
```diff
@@ -306,14 +306,13 @@ if (BUILD_RDIFF)
       src/rdiff.c
       src/isprefix.c)
 
-  add_executable(rdiff ${rdiff_SRCS})
   if (POPT_FOUND)
+    add_executable(rdiff ${rdiff_SRCS})
```
gulikoza: Isn't this already handled with BUILD_RDIFF and set(LAST_TARGET rsync) ?

sgh:
It turns out that fmemopen is not available on windows (which is a platform we target as well as linux). So possibly a windows-way of doing it is to create tmpfiles for it.

Regarding rdif it might be that rdiff can be turned off explicitly. I do however think that autodetecting stuff is beneficial.

I'll issue new pr's for each of the topics and take it from there. So I'll close the pr. Thank you for helping me.

PR #79, opened 2016-07-18
faster hash table using open addressing

MartinNowak:
- replace fixed size hash table with a fitted open addressing one
- size of table is next_pow_2(count / load_factor)
- collisions are resolved w/ a variant of quadratic probing using triangular numbers
- mix 32 bits of rolling adler32 sum for much fewer collisions

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -2,6 +2,7 @@
  *
  * librsync -- the library for network deltas
  *
+ * Copyright (C) 2016 by Martin Nowak <code@dawg.eu>
  * Copyright (C) 1999, 2000, 2001, 2014 by Martin Pool <mbp@sourcefrog.net>
  * Copyright (C) 1999 by Andrew Tridgell
  *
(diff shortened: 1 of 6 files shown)


src/search.c
```diff
@@ -45,112 +46,113 @@
...
-    t1->t = t2->t;
-    t2->t = ts;
-
-    int ti = t1->i;
-    t1->i = t2->i;
-    t2->i = ti;
+static unsigned int next_pow_2(unsigned int v)
+{
+    if (v < 2) return v;
+    return (1U << (32 - __builtin_clz(v - 1)));
```
MartinNowak: Need to port this to other compilers.

src/sumset.h
```diff
@@ -20,54 +21,37 @@
...
+ */
+typedef struct rs_block_sig {
+    rs_weak_sum_t   weak_sum;	/* simple checksum */
+    rs_strong_sum_t strong_sum;	/* checksum  */
+} rs_block_sig_t;
 
 /*
  * This structure describes all the sums generated for an instance of
  * a file.  It incorporates some redundancy to make it easier to
  * search.
```
sourcefrog: Can you add some description here of the new structure of the hash table?

src/search.c
```diff
@@ -45,112 +46,113 @@
...
-    for (n = sums->count - 1; n > 0;) {
-        swap(&sums->targets[0], &sums->targets[n]);
-        --n;
-        for (i = 0; ((i << 1) + 1) <= n;) {
-            k = (i << 1) + 1;
-            if ((k + 1 <= n) && (rs_compare_targets(&sums->targets[k], &sums->targets[k + 1], sums) < 0))
-                k = k + 1;
-            if (rs_compare_targets(&sums->targets[k], &sums->targets[i], sums) > 0)
-                swap(&sums->targets[k], &sums->targets[i]);
+static unsigned int lookup_entry(const rs_signature_t *sigs, rs_weak_sum_t weak_sum, const rs_byte_t *inbuf, size_t block_len)
```
sourcefrog: Can you add a comment, especially as to what it returns.

src/search.c
sourcefrog: Yep, we should do that before merging.

src/search.c
```diff
@@ -45,112 +46,113 @@
...
-    if (v != 0)
-        return v;
-
-    return memcmp(sums->block_sigs[t1->i].strong_sum,
-            sums->block_sigs[t2->i].strong_sum,
-            sums->strong_sum_len);
+// mix full 32-bits of bad adler32 hash before truncating
+static unsigned int mix32(unsigned int h)
+{
+    // final mix function of MurmurHash2
```
sourcefrog: It'd be good to be more clear here what relation this has to the adler hash, if any - I think nothing, this is a separate hash just for the table?

src/search.c
```diff
@@ -45,112 +46,113 @@
...
-    unsigned int i, j, n, k, p;
-    for (i = 1; i < sums->count; ++i) {
-        for (j = i; j > 0;) {
-            p = (j - 1) >> 1;
-            if (rs_compare_targets(&sums->targets[j], &sums->targets[p], sums) > 0)
-                swap(&sums->targets[j], &sums->targets[p]);
-            else
-                break;
-            j = p;
+static bool insert_entry(rs_signature_t *sigs, const rs_block_sig_t *block_sig, unsigned int strong_idx)
```
sourcefrog: Please comment the function.

sourcefrog:
That looks good, thanks, though I could review it better with a bit more description what's going on.

Please add a news entry.

Separately from this pr, I'd like to have a better benchmark in-tree - do you have one you're running?

dbaarda:
FWIW, I had a glance at this and saw the "mix 32 bits of rolling adler32 sum for much fewer collisions" bit that got me thinking. Why is the rollsum not good enough for a hashtable index by itself? Doesn't this indicate it's not a good hash?

So I revisited some earlier conversations about the rollsum quality, did some more thinking, and put together this rough and possibly flawed analysis and recommendations;

http://minkirri.apana.org.au/wiki/LibrsyncRollsumAnalysis

Note this includes suggestions for cheaper rollsum -> hashtable index algorithms that could be applied to this change.

I actually messed around with a bunch other ideas and researched integer hashing algorithms that turned into dead-ends that I didn't include in this analysis, but feel free to ask if you think I've missed something.

MartinNowak:
> So I revisited some earlier conversations about the rollsum quality, did some more thinking, and put together this rough and possibly flawed analysis and recommendations;
> 
> http://minkirri.apana.org.au/wiki/LibrsyncRollsumAnalysis

Looks interesting, but let's keep that out of this PR. Right now the mix function ensures a nice exponential distribution of collision lengths. If an equally well distribution can be achieved cheaper w/ a different rolling hash, then that would be a great followup PR.

dbaarda:
> Looks interesting, but let's keep that out of this PR. Right now the mix function ensures a nice exponential distribution of collision lengths. If an equally well distribution can be achieved cheaper w/ a different rolling hash, then that would be a great followup PR.

There was one bit in that analysis that was possibly related to this PR. Even with the current rollsum, there should be a slightly simpler way to get a hashtable index from the weaksum than the mix32() that should be at least as good.

Instead of doing this;
```C
    const unsigned int mask = sigs->bucket_len - 1;
    index = mix32(weak_sum) & mask;
```
you can just use a mod instead of mix32 like this;
```C
    const unsigned int mask = sigs->bucket_len - 1;
    index = weak_sum % mask;
```
Since mod can be relatively expensive on some platforms, a cheaper way to calculate mod (2^N -1) this is;
```C
    const unsigned int mask = sigs->bucket_len - 1;
    const unsigned int shift = __builtin_ctz(sigs->bucket_len);

    index = weak_sum;
    while (index > mask)
      index = (index >> shift) + (index & mask);
```

MartinNowak:
> Instead of doing this;
> 
>     const unsigned int mask = sigs->bucket_len - 1;
>     index = mix32(weak_sum) & mask;

That is mod(2^N)

> you can just use a mod instead of mix32 like this;
> 
>     const unsigned int mask = sigs->bucket_len - 1;
>     index = weak_sum % mask;

Why should I use mod(2^N - 1)? Doesn't make sense to me, one slot remained unused, and the adler32 hash remains as crappy.
The point of the mix32 function is to use the bits from above 2^N to increase the entropy in the 0..2^N range, this is the part that a good hash would do by itself.

dbaarda:
> Why should I use mod(2^N - 1)? Doesn't make sense to me, one slot remained unused, and the adler32 hash remains as crappy.

It's not actually an adler32... it's a kind of hybrid between adler32 and fletcher32. My analysis suggests that it's crappy in the lower 16 bits, particularly for small block sizes (<192kB), but the upper 16bits should be pretty good. A naive mod(2^N) of weaksum for a hashtable index would keep the crappy low bits and throw away the good high bits.

A mod(2^N-1) on the other hand mixes all the upper bits into it, so it uses all 32bits of the weak sum to give you an Nbit hashtable index.

> The point of the mix32 function is to use the bits from above 2^N to increase the entropy in the 0..2^N range, this is the part that a good hash would do by itself.

That's exactly what mod(2^N-1) does, and possibly even better than the mix32 does, at the expense of leaving one slot unused (though it could still be used for collisions).

The mix32 integer hash attempts to mix the information in all 32bits across each other, so each bit in the 32bit result contains information from a mix of many input bits. However, the mod(2^N) then throws away the high bits, possibly throwing away some of the information that was mixed into those high bits.

A mod(2^N-1) mixes all 32bits down into N bits, discarding nothing. This is what Fletcher32 uses when accumulating the 16 bit s1 and s2 sums (but librsync doesn't), and is equivalent to using 1's compliment addition. The effect is that every thing that overflows past the N'th bit gets end-over-carried back into the least significant bit. This mixes everything past the Nth bit back into the lower N bits.

MartinNowak:
> That's exactly what mod(2^N-1) does, and possibly even better than the mix32 does, at the expense of leaving one slot unused (though it could still be used for collisions).

Well I got a perfect exponential distribution of the collisions, so it can't be that bad.

> A mod(2^N-1) mixes all 32bits down into N bits, discarding nothing.

Any reference for that, looks indeed like an interesting property. But your loop still looks too expensive. Also let's end that OT topic discussion, the mix32 function does it's job, and if you a better alternative, we can implement that later on.

MartinNowak:
I'm struggling a bit to come up w/ a good benchmark. The synthetic files that I tried to generate aren't representative for real world usages. Any idea for a good benchmark?
We could indeed try to bench delta computation for some .qcow images and .sql.gz dumps. But where to get the data?

MartinNowak:
Mmh, how about https://dumps.wikimedia.org/other/pagecounts-raw/2016/2016-01/ (found here [Sample datasets for benchmarking and testing](https://www.percona.com/blog/2011/02/01/sample-datasets-for-benchmarking-and-testing/))?

dimsua:
Hello
I am testing ploop-image 88GB with your rdiff-patches.
1. Make signature for 88GB ploop image
2. Write 1.1GB random data in ploop image
3. Make delta
fast-rdiff: real 8m58.990s
rdiff: real **32m**47.387s

MartinNowak:
Have you used the same block size? Fast-diff uses 1MB blocks while rdiff can be configured but uses 2KB blocks by default. You should also increase rdiff's input and output buffers, b/c they're fairly small by default which causes a lot of I/O overhead.

dbaarda:
FWIW I thought that it's pretty sad that we have to write our own hashtable, and surely there must be generic hashtable implementations we could use. After looking around I came to the conclusion most of them are either way too complicated or way too limited for our needs.

So I wrote my own generic hashtable that we could use if we wanted. It's pretty well documented and tested. I've got it in my dev/hashtable1 branch;

https://github.com/dbaarda/librsync/blob/dev/hashtable1/src/hashtable.h

dbaarda:
Again, incase anyone is interested in testing it, I've updated my dev/hashtable branch to actually use the generic hashtable I wrote, and also merged in all the stuff in my tidy/signature1 pull request;

https://github.com/dbaarda/librsync/tree/dev/hashtable1

It already runs the existing tests a tiny bit faster, but the real benefits should only show up for very large files.

I'm also contemplating adding support for rsync heuristics for recommended block sizes and strong sum lengths based on basis file size. This should make a huge difference for large files, because the default 2K block size is terrible for them (rsync uses block_len=sqrtt(basis_fsize)).

dbaarda:
MartinNowak: I apologize for questioning your mix32() function.

After doing more testing I've adopted the MurmurHash3 finalizer in my hashtable.c experiments too. I compared the MurmurHash2 and MurmurHash3 finalizers and the v3 one performed a bit better. Even though the v3 one is a bit more overhead to calculate, the better distribution it gets results in less clustering and less compares, which ends up cheaper overall.

The Rollsum has nasty clustering, particularly for small block sizes. For 1K blocksize  you end up using only the lower 14 bits of the s1 sum, which means they are all clustered together in 1/4 of the available s1 space.

Using "% mask" instead of "& mask" as I proposed helps a fair bit, but nowhere near as well as as MurmurHash's finalizers for hashtables larger than 2^16.  The problem is the "end-over-carry" that "% mask" does mixes the truncated s2 high bits into the s1 low bits, but it is the high s1 bits that need the mixing, so it still clusters.

MurmurHash's finalizers are magic. I thought I might be able to invent something that takes advantage of the fact we are reducing the number of bits for the index, but I soon discovered this stuff is mysterious magic. Things like murmurhash are not designed, they are brute forced to find the right combination of mysterious constants.

dbaarda:
On testing: my hashtable experiments have added a "largefile.test" that generates a 1G old/new file pair which has 50% matches, 50% misses, and then does the sig delta patch using a blocksize of 1K. This is pretty brutal on the search function so it's a good hashtable test.

MartinNowak:
Back from vacation now. My last state was a similar test using ubuntu-16.04-desktop-amd64.iso ubuntu-16.04.1-desktop-amd64.iso, those had a similar 50/50 match/miss rate.
With those tests I couldn't really produce a speedup over the existing hash table due to many cache misses and failing branch predictions. I need to properly retest/profile and then we might have to optimize the memory size of the hash table a bit.

> On testing: my hashtable experiments have added a "largefile.test" that generates a 1G old/new file pair which has 50% matches, 50% misses, and then does the sig delta patch using a blocksize of 1K.

Nice, will try that.

> I'm also contemplating adding support for rsync heuristics for recommended block sizes and strong sum lengths based on basis file size.

Yes, this is a good idea, but we should keep that separate of the hash table change.

dbaarda:
I've been doing a bunch of experiments and benchmarking and am approaching the point of diminishing returns in my dev/hashtable1 branch.

I've added a performance.test test that excecises largefile.test for a bunch of "standard" block numbers to exercise different load factors etc.

I learned a hell of a lot from your push request about hashtables that I was initially getting quite wrong. I started with a simple hashtable that just contained pointers to block_sigs, but the performance was terrible. This made me realize just how important memory locality is on modern CPU's. Next I went to buckets containing weaksum&pointer to blocksigs... it was better but still not as good as yours, which I couldn't understand at first. Then I realized on 64bit platforms pointers are 8 bytes, but integer indexes are only 4, so you were getting better "bucket packing" As a quick test I tried buckets of weaksum&index into array of pointers to block_sigs and it got better even though  it added an extra layer of indirection! After that I switched to my current buckets containing only weaksums and another array with the same index of pointers to block_sigs.

The other thing to take into account is librsync's usage pattern tends to have a huge number of "lookup misses" compared to "lookup hits", so it's worth optimizing for misses. For example, for a hashtable with 1M * 1K blocks doing a match with 50/50 hits/misses and loadfactor=0.5, we get the following match stats;

librsync: match statistics: signature[537395200 searches, 524288 (0.098%) matches, 627469786 (1.168x) weak sum compares, 1548850 (0.288%) strong sum compares, 1547240 (0.288%) strong sum calcs]

For 716K * 1K blocks and loadfactor 0.7 we get;

librsync: match statistics: signature[375756800 searches, 366592 (0.098%) matches, 1053222017 (2.803x) weak sum compares, 867330 (0.231%) strong sum compares, 866681 (0.231%) strong sum calcs]

The performance difference between my dev/hashtable1 and your faster_hashtable is now pretty minimal. Mine tends to win on higher loadfactors because of better bucket locality for bucket collisions/misses (just weaksums), and yours is a tiny bit better for low loadfactors and huge numbers of blocks because it has slightly better locality for strongsum compares (which we do slightly more of for weaksum collisions). The main other differences is; yours is a smaller diff, mine is maybe a bit neater.

The other thing I've found is test timing can vary a fair bit for very little reason, and the results might be very different for different platforms. I'm using a pretty crappy Intel(R) Atom(TM) CPU D525@1.80GHz which helps highlight when things are slow, but better CPU's with larger caches might have completely different optimization behaviours.

dbaarda:
FWIW I've put up https://github.com/librsync/librsync/pull/95 as an alternative implementation of this.

dbaarda:
I've also done the merge against master and resolved all the conflicts in my faster_hashtable branch, incase anyone wants to pursue this implementation further. It was a little messy after the signature tidy stuff I did. It's definitely a much smaller diff than my implementation.

MartinNowak:
Closed in favor of #95.

PR #80, opened 2016-07-20
[RFC] Implement optional full block matching

gulikoza:
Hi,

librsync at the moment uses rolling checksum to compare and find each possible matching block for the given signature. This is great for files which change and move existing content (mbox mailboxes for example) and provides the most optimal delta (given limited bandwidth for example). However, for some other content, this is rarely useful. It's highly unlikely that filesystem image (virtual machine disk) for example will move existing block to a different location. Even if signature block size would match filesystem block size, it's unlikely that a block will be copied exactly to some different location within the filesystem without a single byte changed.

For files that store content in blocks that don't arbitrarily move within the file, the rolling checksum does not provide any advantage and only increases the load (and collision possibility) by check-summing each and every byte. It would therefore better to allow only full (sig size) block matching in those cases. Virtual machine disk images for example can be huge. Storing deltas for such files (backups for example) usually happens over LAN, so optimizing local I/O and CPU usage to speed up processing is preferable over getting a slightly smaller delta (if it's smaller at all) and conserving bandwidth.

Block match is activated with -m parameter. When it's active and the current block is not found in the signature, entire sig-size block will be marked not-matching (rs_appendmiss) and sums will be re-initialized from the next signature block. This is v2 of the patch, I had a v1 where the block match size would not have to be the same as signature block size (for instance, searching a file by moving the checksum window 512-bytes in a 2048-byte blocks), but the code was much more complicated. Having the window only move in signature-size blocks is much more simple and arbitrary size would hardly provide any benefit in this case (if it does, then normal rolling checksum mode can be used).

By moving the search window in sig-size blocks allows for additional improvements, not possible (or significantly harder to implement) with the current algorithm:
- we can mach exactly the same block in the source file first, before searching for other matching blocks (optimizes delta - included in the second part of the PR)
- patching could be improved to support in-line patching (w/o copying the destination file)
- an inverse-delta could be created at the same time, while doing an inline patch of the destination file
- ...

The usage case for this mode would be any file that stores content in blocks which don't move. By definition these are usually larger files, since moving existing blocks is I/O intensive. By carefully choosing signature block size (to match content block size closely) and using block match mode when generating delta, the user can improve delta processing speed.

Would something like this be worth considering?

### DIFF
--- a/src/delta.c
+++ b/src/delta.c
@@ -180,6 +180,10 @@ static rs_result rs_delta_s_scan(rs_job_t *job)
                 RollsumInit(&job->weak_sum);
             }
         }
+
+        if (job->block_match > 0) {
+            job->block_match++;
+        }
     }
     /* if we completed OK */
     if (result==RS_DONE) {
@@ -207,6 +211,8 @@ static rs_result rs_delta_s_flush(rs_job_t *job)
     rs_getinput(job);
     /* output any pending output */
     result=rs_tube_catchup(job);
+    /* tokens should no longer be valid */
+    job->block_match = -1;
     /* while output is not blocked and there is any remaining data */
     while ((result==RS_DONE) && (job->scoop_pos < job->scoop_avail)) {
         /* check if this block matches */
@@ -280,7 +286,7 @@ inline int rs_findmatch(rs_job_t *job, rs_long_t *match_pos, size_t *match_len) 
                                *match_len,
                                job->signature,
                                &job->stats,
-                               match_pos);
+                               match_pos, job->block_match);
 }
 
 
--- a/src/delta.c
+++ b/src/delta.c
@@ -159,20 +159,25 @@ static rs_result rs_delta_s_scan(rs_job_t *job)
             result=rs_appendmatch(job,match_pos,match_len);
             RollsumInit(&job->weak_sum);
         } else {
-            /* rotate the weak_sum and append the miss byte */
-            RollsumRotate(&job->weak_sum,job->scoop_next[job->scoop_pos],
+            if (!job->block_match) {
+                /* rotate the weak_sum and append the miss byte */
+                RollsumRotate(&job->weak_sum,job->scoop_next[job->scoop_pos],
                           job->scoop_next[job->scoop_pos+job->block_len]);
-            result=rs_appendmiss(job,1);
-            if (rs_roll_paranoia) {
-                RollsumInit(&test);
-                RollsumUpdate(&test, job->scoop_next+job->scoop_pos,
+                result=rs_appendmiss(job,1);
+                if (rs_roll_paranoia) {
+                    RollsumInit(&test);
+                    RollsumUpdate(&test, job->scoop_next+job->scoop_pos,
                               job->block_len);
-                if (RollsumDigest(&test) != RollsumDigest(&job->weak_sum)) {
-                    rs_fatal("mismatch between rolled sum %#x and check %#x",
+                    if (RollsumDigest(&test) != RollsumDigest(&job->weak_sum)) {
+                        rs_fatal("mismatch between rolled sum %#x and check %#x",
                              (int)RollsumDigest(&job->weak_sum),
                              (int)RollsumDigest(&test));
+                    }
                 }
-                
+            } else {
+                /* append block_len miss bytes */
+                result=rs_appendmiss(job,job->block_len);
+                RollsumInit(&job->weak_sum);
             }
         }
     }
@@ -440,7 +445,7 @@ static rs_result rs_delta_s_header(rs_job_t *job)
 }
 
 
-rs_job_t *rs_delta_begin(rs_signature_t *sig)
+rs_job_t *rs_delta_begin(rs_signature_t *sig, int block_match)
 {
     /* Caller must have called rs_build_hash_table() by now */
     if (!sig->tag_table)
@@ -450,6 +455,7 @@ rs_job_t *rs_delta_begin(rs_signature_t *sig)
 
     job = rs_job_new("delta", rs_delta_s_header);
     job->signature = sig;
+    job->block_match = block_match;
 
     RollsumInit(&job->weak_sum);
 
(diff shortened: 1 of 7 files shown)


MartinNowak:
> I had a v1 where the block match size would not have to be the same as signature block size (for instance, searching a file by moving the checksum window 512-bytes in a 2048-byte blocks), but the code was much more complicated.

This shouldn't be more complicated but really simple. Just feed some more bytes to the rolling checksum before trying to match. To make this simple you could enforce that the given block-size is a multiple of the match-stride.

src/search.c
```diff
@@ -178,10 +178,21 @@ rs_search_for_block(rs_weak_sum_t weak_sum,
...
     int r = bucket->r + 1; /* right bound of search region */
     int v = 1; /* direction of next move: -ve left, +ve right */
+    int i = -1;
 
     if (l == NULL_TAG)
         return 0;
 
+    /* Try a direct match */
+    if ((job_token > 0) && (job_token < sig->count) &&
+        (weak_sum == sig->block_sigs[job_token-1].weak_sum)) {
```
MartinNowak: This is an ugly hack around a slow hash table, see #79.

src/delta.c
```diff
@@ -159,22 +159,31 @@ static rs_result rs_delta_s_scan(rs_job_t *job)
...
+                        rs_fatal("mismatch between rolled sum %#x and check %#x",
                              (int)RollsumDigest(&job->weak_sum),
                              (int)RollsumDigest(&test));
+                    }
                 }
-                
+            } else {
+                /* append block_len miss bytes */
+                result=rs_appendmiss(job,job->block_len);
+                RollsumInit(&job->weak_sum);
```
MartinNowak: Where is the weak_sum actually computed? Only RollsumInit seems to get called here.

MartinNowak:
Maybe I missed sth. but are you actually still computing a weak hash? It seems to me that you're only relying on exact block matches at exactly the same position.

> Even if signature block size would match filesystem block size, it's unlikely that a block will be copied exactly to some different location within the filesystem without a single byte changed.

Have at least verified this hypothesis? I find it hard to believe that a filesystem, e.g. ext4, stores changing metadata in every 4kB block.
I think they're using optimized allocators in special blocks that manage a huge amount of 4kB blocks. And I'd think that defragmenting/compacting existing blocks would be a reasonable operation for a filesystem.
Regarding the slow performance of librsync with many small blocks, please give #79 a try, it should speed up delta computation by 4-5x for a 4GB file with 2kB blocksize.

src/delta.c
gulikoza: weak_sum is calculated in rs_findmatch:

    /* calculate the weak_sum if we don't have one */
    if (job->weak_sum.count == 0)

It is reset and then calculated for the entire block on the next pass.

gulikoza:
Full blocks are still matched if the signature block size matches. So if you do a 4KB signature block size, matching moved blocks will be found, exactly as they are now, this patch does not change that. What is unlikely is that a 4KB ext4 block would be moved 2KB, or 1KB or a few bytes up or down. Searching for these blocks using RollingChecksum seems pointless as defragmenting won't do that :) Even images like qcow2 are allocated in 64K blocks AFAIK.

I saw #79 and plan to test it and this is also one of the reasons I opened this PR. I had this prepared for some time, but having more options (or doing a combination of both) would maybe produce the best results...

src/search.c
gulikoza: Maybe, but testing first if the same block in the source file matches seems to be a good guess. If a file has multiple identical blocks, which one do you take as a reference? The first one, the last one...depends on the implementation of the hash table. Since they are identical it doesn't matter where you take the data (it's the same), but delta file will be smaller if blocks can be merged together (so you get only a single operation spanning multiple blocks).

gulikoza:
> This shouldn't be more complicated but really simple. Just feed some more bytes to the rolling checksum before trying to match. To make this simple you could enforce that the given block-size is a multiple of the match-stride.

I don't know, maybe the it was because of the first try and the code was more complicated...but, in this version the weak_sum is only re-inited and the existing code handles the rest. Before I had to feed bytes to the rolling checksum, check when the block boundary was hit and so on... as said before, choosing a smaller (or more fitting) signature block size might be more beneficial.

Thanks for all the comments so far!

src/search.c
MartinNowak: Fair enough. In the new hashtable I omit duplicate blocks, b/c only the first one can ever be matched.

src/delta.c
MartinNowak: Yuck, why would one spread ingesting the data over multiple functions?

MartinNowak:
> Full blocks are still matched if the signature block size matches. So if you do a 4KB signature block size, matching moved blocks will be found, exactly as they are now, this patch does not change that. 

OK, the way it's stiched together, I didn't expect findmatch to actually compute a hash sum.

> What is unlikely is that a 4KB ext4 block would be moved 2KB, or 1KB or a few bytes up or down. 

Yes true, but maybe other people have use cases to match 64KiB blocks that get shifted by 8KiB inserts or so (think of db images or so). So if we can easily make this work with any match-stride it might be a bit more flexible. I'll give it a try in the next few days if I can find some time.

sourcefrog:
The general idea of matching on only whole blocks sounds great: it will also address some of the risks of mischief from malicious data that @therealmik has previously identified.

dbaarda:
@gulikoza In the interests of tidying up the outstanding pull requests, I've had another look at this.

This pull request was submitted in 2016 against v2.2.0, before the new hashtable implementation in v2.0.1 which significantly sped things up. There have also been additional speedups since then, and a delta on a 1GB file that used to take more than 7min now takes less than 1min (See https://github.com/dbaarda/librsync-tests/blob/master/RESULTS.rst). This pull request also no longer applies cleanly against the code.

As @MartinNowak says, this pull request was largely an ugly hack around a slow hashtable, which has now been fixed. It is true that there are some file types that use fixed internal block sizes that delimit changes, like filesystem images, and librsync could be modified to take advantages of that to calculate delta's even faster. However, I think it would be better to do that using an "atom-size" independent of the block-size (which would be the number of "atoms" per block, where the default atom-size is one byte), so you could eg roll over a file of atom-size=32 byte records using a block-size=16 records per block. This would speed things up, but the benefits are much less now the hashtable is faster, and it's unclear how much demand there would be for this.

Given all of the above, I'm going to abandon this pull request. If you still think this is worth doing, feel free to submit a new pull request against the current code that implements an "atom-size" approach.

PR #83, opened 2016-07-31
Don't call realloc for every block when reading the signature file.

gulikoza:
This is the last bit from the #67.
Calling realloc for every signature block in Mingw is extremely slow for some reason, reading 2M blocks can take 2 and a half minutes. Optimize by pre-allocating the needed memory.

The algorithm will still fallback to realloc if the user supplies a wrong (too low) count for job->estimated_signature_count. File size detection brings in fstat (as per #67 comments) which is unfortunately somewhat more difficult (uses fstat, fstat64 or _fstati64 if available). The code is clearer, only 1 if() is added in rs_loadsig_add_sum() compared to previous version.


sourcefrog:
Seems good, thanks!

dbaarda:
FTR, there is a minor issue with this pull request that I've encountered.

The job_t type is opaque from the librsync-dev perspective, where only things in librsync.h, librsync-config.h and types.h are visible.

This means people wanting to use a precompiled librsync-dev package for implementing their own streaming stuff can never set sig_file_bytes or estimated_signature_count in the job.

It's unclear exactly how best to fix this, but IMHO these are the options;

1) Change the librsync API and add either a block_num or sigfile_size argument to rs_loadsig_begin() as an optional hint. Note if we are going to do this it might be nice to add an oldfile_size hint to rs_sig_begin() that it can use to set default block_len and strong_len values from the filesize using the same kind of heuristic rsync uses; 

    block_len=sqrt(filesize), 
    strong_bits=(E + 2*log2(filesize) - log2(block_len))

where probability of corruption ignoring extra security from the rollsum is less than 1/2^E. This would be a non-backwards compatible API change.

2) Keep the API backwards compatible and add an extra method for injecting the filesize hint to the job after it's been created by rs_loadsig_begin(). This technique might also be usable for injecting filesize hints to rs_sig_begin() in a backwards compatible way, though maybe a better way would be to to add a helper function to calculate the recommended values from a filesize for that.

3) Don't make the filesize hint publicly available for the streaming API and only use it under the hood for the whole file API. Note that whole.c could in theory be implemented using only the public streaming API but isn't; it bypasses the API to directly manipulate the job for this feature, and also to grab the statistics instead of using the public API rs_job_statistics().

4) Drop the filesize hint and instead go with O(ln(N)) reallocs using a "start with 16 and double size when full" approach. It can also "realloc back down to used size" after the signature is fully loaded. I have work in progress that was adding this when there was no filesize hint, but maybe we should just use that instead of the filesize hint.

sourcefrog:
We could add a new API that has the option - I don't see a reason to break existing callers. 

Knowing the length of the file is probably common enough that we shouldn't need to do multiple reallocs.

PR #84, opened 2016-08-01
Fix build with VS2015 Win64 - suppress warning about size_t vs sizeof(long)

pol51:
#warning preprocessor directive doesn't exists on MSVC

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,13 @@ if (NOT CMAKE_SYSTEM_NAME)
 	message(FATAL_ERROR "No target OS set")
 endif()
 
+set(DO_RS_TRACE 0)
+option(ENABLE_TRACE "Compile in detailed trace messages" OFF)
+if (ENABLE_TRACE)
+    set(DO_RS_TRACE 1)
+endif (ENABLE_TRACE)
+message(STATUS "DO_RS_TRACE=${DO_RS_TRACE}")
+
 # Add an option to include compression support
 option(ENABLE_COMPRESSION "Whether or not to build with compression support" OFF)
 # TODO: Remove this warning when compression is implemented.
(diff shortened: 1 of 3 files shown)


sourcefrog:
Is this condition true on that platform?

How about if we make it an error instead?

pol51:
Yes, this condition it was true on this platform.

If you make it an error it's like you're not supporting this platform anymore (like now in fact as #warning directive is an error for MSVC).

sourcefrog:
Suppressing the message isn't helping anyone; let's make sure to use a printf format that handles `size_t` correctly instead. In C99 I think that's %zu.  We could add that near 
https://github.com/librsync/librsync/blob/master/src/trace.h#L50

pol51:
Indeed it would be better to fix the real issue.
My point was just to make it non-blocking on VS2016/Win64 like it is with GCC (as a warning doesn't stop the compilation with GCC).

gulikoza:
What would be the drawback of using PRINTF_CAST_U64 and using PRINTF_FORMAT_U64 and cast to rs_long_t instead of casting to long? Long is 32-bit on some 64-bit platforms and %zu was only added in VS2015, previous versions don't support it.

sourcefrog:
> What would be the drawback of using PRINTF_CAST_U64 and using PRINTF_FORMAT_U64 and cast to rs_long_t instead of casting to long? Long is 32-bit on some 64-bit platforms and %zu was only added in VS2015, previous versions don't support it.

That sounds good to me.

dbaarda:
I just saw this again after merging https://github.com/librsync/librsync/pull/94 which removed this warning as "probably obsolete".

It sounds like removing the warning is only part of the fix needed for this. The other recent merge was https://github.com/librsync/librsync/pull/92 which cemented our dependency on C99's stdint.h, which probably makes a lot of this easier on the printf front.

I'll try to find some time to have a look at this in the near future.

gulikoza:
I don't believe Windows fully supports all the printf formats, unfortunately.
I found at least one bug report (https://sourceforge.net/p/mingw/bugs/2279) which would suggest that MSVCRT.DLL does not support %jd.
As much as this 32/64-bit mess should be solved in this day & age, having a custom 64-bit type still seems relevant, even if it's defined to intmax_t on most platforms.

PR #85, opened 2016-08-10
Optimize rollsum.c to add char offsets once.

dbaarda:
The char offsets in the rollsum don't need to be added incrementally, but can be calculated and added once.

### DIFF
--- a/src/rollsum.c
+++ b/src/rollsum.c
@@ -27,9 +27,10 @@
 #define DO8(buf,i)  DO4(buf,i); DO4(buf,i+4);
 #define DO16(buf)   DO8(buf,0); DO8(buf,8);
 
-void RollsumUpdate(Rollsum *sum, const unsigned char *buf, size_t len) {
+void RollsumUpdate(Rollsum *sum, const unsigned char *buf, size_t len)
+{
     /* ANSI C says no overflow for unsigned. zlib's adler32 goes to extra
-    effort to avoid overflow for its mod prime, which we don't have.*/
+       effort to avoid overflow for its mod prime, which we don't have. */
     size_t n = len;
     uint_fast16_t s1 = sum->s1;
     uint_fast16_t s2 = sum->s2;
@@ -46,8 +47,8 @@ void RollsumUpdate(Rollsum *sum, const unsigned char *buf, size_t len) {
     }
     /* Increment s1 and s2 by the amounts added by the char offset. */
     s1 += len * ROLLSUM_CHAR_OFFSET;
-    s2 += ((len*(len+1))/2) * ROLLSUM_CHAR_OFFSET;
-    sum->count+=len;  /* Increment sum count. */
-    sum->s1=s1;
-    sum->s2=s2;
+    s2 += ((len * (len + 1)) / 2) * ROLLSUM_CHAR_OFFSET;
+    sum->count += len;          /* Increment sum count. */
+    sum->s1 = s1;
+    sum->s2 = s2;
 }
--- a/src/rollsum.c
+++ b/src/rollsum.c
@@ -27,12 +27,12 @@
 #define DO8(buf,i)  DO4(buf,i); DO4(buf,i+4);
 #define DO16(buf)   DO8(buf,0); DO8(buf,8);
 
-void RollsumUpdate(Rollsum *sum,const unsigned char *buf,unsigned int len) {
+void RollsumUpdate(Rollsum *sum, const unsigned char *buf, size_t len) {
     /* ANSI C says no overflow for unsigned. zlib's adler32 goes to extra
     effort to avoid overflow for its mod prime, which we don't have.*/
-    unsigned int n = len;
-    unsigned long s1 = sum->s1;
-    unsigned long s2 = sum->s2;
+    size_t n = len;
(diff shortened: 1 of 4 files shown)


src/rollsum.c
```diff
@@ -26,26 +26,28 @@
...
 #define DO8(buf,i)  DO4(buf,i); DO4(buf,i+4);
 #define DO16(buf)   DO8(buf,0); DO8(buf,8);
-#define OF16(off)  {s1 += 16*off; s2 += 136*off;}
 
 void RollsumUpdate(Rollsum *sum,const unsigned char *buf,unsigned int len) {
-    /* ANSI C says no overflow for unsigned. 
-     zlib's adler 32 goes to extra effort to avoid overflow*/
+    /* ANSI C says no overflow for unsigned. zlib's adler32 goes to extra
+    effort to avoid overflow for its mod prime, which we don't have.*/
+    unsigned int n = len;
```
sourcefrog: It'd be better if you left the param `len` unchanged through the function and counted down in `n`.

src/rollsum.c
```diff
@@ -26,26 +26,28 @@
...
     }
     while (len != 0) {
-        s1 += (*buf++ + ROLLSUM_CHAR_OFFSET);
+        s1 += *buf++;
         s2 += s1;
         len--;
     }
+    /* Increment s1 and s2 by the amounts added by the char offset. */
+    s1 += n * ROLLSUM_CHAR_OFFSET;
+    s2 += ((n*(n+1))>>1) * ROLLSUM_CHAR_OFFSET;
```
sourcefrog: If the `>>1` is not specifically because you need some specific shifting behavior for overflow (presumably not given `n` is the length in memory), I think just `/2` would be clearer.

sourcefrog:
If I was writing this today, I would add unit tests for `rollsum.c` checking against specific known values.  How about adding some now so we can be more clear this isn't changing anything?

It looks safe enough though.

src/rollsum.c
dbaarda: Yeah, agree. I mostly did it this way for diff minimization, but changing it now.

src/rollsum.c
dbaarda: I did it for optimization reasons, but I guess every compiler these days is smart enough to just use a shift under the hood. Doing /2 is definitely clearer. Changing it now.

dbaarda:
I'll add some tests in a bit. BTW, what is the best way to write C tests these days? I'm going to start with simple asserts like in isprefix_test.c, but maybe something like https://libcheck.github.io/check/ would be good?

There's another change that I'd like to do at some point but I dunno if this pull request is the right place. I'd like to change s1 and s2 in the Rollsum struct from "unsigned long" to "unsigned int". They only need to be 16bits and this will free the compiler to use whatever is optimal for that. It should help a little on any 16bit platforms, like maybe embedded systems?

Is there any good reasons not to do this?

sourcefrog:
I was just going to add C binaries like `isprefix_test.c` that return 0 or 1.  Use the existing logging to say what's happening.  That fits nicely with CMake and is portable.  I would somewhat avoid adding dependencies.

> There's another change that I'd like to do at some point but I dunno if this pull request is the right place. I'd like to change s1 and s2 in the Rollsum struct from "unsigned long" to "unsigned int". They only need to be 16bits and this will free the compiler to use whatever is optimal for that. It should help a little on any 16bit platforms, like maybe embedded systems?

Let's use `int_fast16_t` for that then, which is precise about what we need.  http://en.cppreference.com/w/c/types/integer

dbaarda:
Whoah! Actually, when writing tests I discovered that on my platform unsigned int is 32 bits and unsigned long is 64! This is way bigger than it needs to be, and means I probably have to do explicit masking to prune the digest down to 32bits or risk things not matching if you use RollsumDigest wrong.

I noticed we have types.h that generates our own uintXX_types if stdint.h is not available, but the blake2 stuff is using stdint.h directly. Do we really need types.h or should we just assume/require that C99 stdint.h is available?

There is code in checksum.c that re-implements RollsumUpdate() and assumes unsigned int is 32 bits. mksum.c also assumes unsigned int is 32 bits. The use of unsigned long in rollsum.c and unsigned int in checksum.c could mean on different platforms we end up with different signatures, or even have signatures that don't match correctly on the same platform.

I think we've been a bit lucky this hasn't bitten us. I think we've got away with it because nearly everything has unsigned int as 32bits, and for the differences in rollsum.c and checksum.c to show up then we'd need to overflow 32 bits in s1... so very very large block sizes.

I'll put together another change to try and tidy some of this up.

dbaarda:
Looking closer again, the differences in types used by rollsum.c and checksum.c would only cause problems on platforms where unsigned int is less than 32bits, so we are safe.

But it's still messy...

sourcefrog:
> Whoah! Actually, when writing tests I discovered that on my platform unsigned int is 32 bits and unsigned long is 64! 

tests hey 😄 

1999 is a long time ago...  It seems like now it'd be reasonable to rely on C99 features like `stdint.h` and just provide our own workarounds, under the same name, on platforms where it's not fully implemented.

dbaarda:
I'm pretty happy with this now, and think it's ready to merge.

PR #88, opened 2016-09-01
Tidy signature handling code.

dbaarda:
This makes `rs_signature_t` more like a class with methods for interacting with it.

This removes some redundant code that duplicated things like signature sanity checking and strongsum algorithm selection based on the signature magic. This makes the sanity checking way more consistent and will make it easier to add different checksum and rollsum algorithms or hashtable implementations.

Removes `job->estimated_signature_count` and renames `job->sig_file_bytes` to `job->sig_fsize`. The estimated block count is is redundant with file size for estimating the number of blocks to pre-allocate. Moves the calculation of block count into `rs_signature_init()` method which takes an `sig_fsize` argument. Note this is also not exposed at the API, so can only be used for whole file operations.

Changes the signature block allocator to fall back when `sig_fsize` is not available to O(ln(N)) allocations instead of O(N) by doubling the allocated size each time. This should speed up applications loading signatures using the streaming API which cannot provide `job->sig_fsize`.

Adds `rs_realloc()` function to `util.h` that does realloc or die to match rs_alloc(). This is used by the signature for resizing the allocated block_sums.

Adds `job->job_owns_sig` attribute used to indicate when the signature should be destroyed with the job. This is used by mksum.c which creates a signature but never passes it to the client. Note changing mksum.c to return the signature to the client would enable direct basis->delta calculations without having to save/load the signature.

Removes job's redundant `magic`, `block_len`, and `strong_sum_len` attributes and makes everything use the signature's equivalent attributes instead. Replaces them in job with `sig_magic`, `sig_block_len`, and `sig_strong_len` attributes that are used only for initializing the signature in mksum.c and loadsums.c. The new names are to avoid confusion between the job and signatures attributes.

Fixed "slack deltas" support in delta.c so that slack deltas are used when doing a delta operation with a NULL signature. Tidied up the code supporting slack deltas a bit.

It makes all of the job type's workflows consistent so that `rs_*_begin()` operations always return a valid `rs_job_t` instance and can never fail except by assertions that should never happen in a valid program. Failures instead are always reported via iteration callbacks returning error results. Note that previously some `rs_*_begin()` operations had fatal errors or could return `NULL`, but we were not checking for this where we call them in `whole.c`.

I also did a bit of indent tidying along the way that I decided to revert at the end to minimize the final diffs. Sorry for the commit noise this causes. I think we should standardize on a reformatting tool and style (indent? clang-format?) and clean it all up, but that's something for another pull request.

### DIFF
--- a/src/delta.c
+++ b/src/delta.c
@@ -79,9 +79,6 @@
 #include "types.h"
 #include "rollsum.h"
 
-const int RS_MD4_SUM_LENGTH = 16;
-const int RS_BLAKE2_SUM_LENGTH = 32;
-
 /**
  * 2002-06-26: Donovan Baarda
  *
--- a/src/delta.c
+++ b/src/delta.c
@@ -21,10 +21,11 @@
  */
 
                                 /*
-                                   | Let's climb to the TOP of that
-                                   | MOUNTAIN and think about STRIP
-                                   | MINING!!
+                                 | Let's climb to the TOP of that
+                                 | MOUNTAIN and think about STRIP
+                                 | MINING!!
                                  */
+
 
 /*
  * delta.c -- Generate in streaming mode an rsync delta given a set of
@@ -59,6 +60,7 @@
  *
  * If nothing matches, it is not so good.
  */
+
 
 #include "config.h"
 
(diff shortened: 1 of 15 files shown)


dbaarda:
Oh yeah, I also fixed "slack deltas"... calculating a delta without a signature, giving you a whole-file delta.

The code was there but had been severed from accessibility by changes along the way.

dbaarda:
Hold on a sec before approving this... I have a few more tidyups on the way for this.

I'll also add tests for the rs_signature_t methods.

dbaarda:
I'm pretty happy with the code now, only thing left to do is add tests for signatures.

dbaarda:
This is now fully complete with included tests.

I'm using it as a crucial improvement to the rs_signature_t handling code that makes it much easier to experiment with different hashtable and rollsum implementations.

src/sumset.h
```diff
@@ -1,73 +1,107 @@
...
+    int count;                  /**> Total number of blocks. */
+    int size;                   /**> Total number of blocks allocated. */
+    rs_block_sig_t *block_sigs; /**> The block signatures for all blocks. */
+    rs_tag_table_entry_t *tag_table;
+    rs_target_t *targets;
 };
 
+/** Initialize an rs_signature instance.
+ *
+ * /param *sig the signature to initialize.
```
sourcefrog: Should be `\param` all through here

src/delta.c
```diff
@@ -140,6 +137,7 @@ static inline rs_result rs_processmiss(rs_job_t *job);
  * scoop and input buffer. */
 static rs_result rs_delta_s_scan(rs_job_t *job)
 {
+    size_t         block_len = job->signature->block_len;
```
dbaarda: Done.

src/delta.c
```diff
@@ -256,12 +253,14 @@ void rs_getinput(rs_job_t *job) {
  * decrementing scoop_pos as appropriate.
  */
 inline int rs_findmatch(rs_job_t *job, rs_long_t *match_pos, size_t *match_len) {
+    size_t block_len = job->signature->block_len;
```
dbaarda: Done

src/sumset.h
dbaarda: Done.

dbaarda:
Ping? I think this is ready. I fixed the doxygen \param's, added a few const's, and fixed a few other doxygen comment errors I found.

I also fixed tests/triple.test touse $bindir so it works with cmake alternate build directories.

Is there anything else you'd like me to do with this before it can be merged?

PR #89, opened 2016-09-10
Fix mutate.test to actually use the hashopt args.

dbaarda:
A tiny deficiency in one of the tests I noticed... it's not actually excercising all the different hashopts it's supposed to.


sourcefrog:
! 😞

PR #91, opened 2016-10-22
Move types.h inline in librsync.h

### LINKED ISSUE #90: types.h not found
As I've been having this `rs_search_for_block` crash problem ( #50 ) on 2.0.0 occasionally, I gave the master branch a shot. When I was compiling against the newly installed librsync, a compilation error pops out:

> /usr/local/include/librsync.h:40:19: fatal error: types.h: No such file or directory

I can see that `librsync.h` is including a `types.h`, but there's no sight of this file under `/usr/local/include`. Maybe there's one step missing in installation process to copy this file?

sourcefrog:
Avoids errors building from an installed copy of librsync.h, when
types.h cannot be found.

Fixes #90.


src/librsync.h
```diff
@@ -37,9 +37,36 @@
 #define _RSYNC_H
 
 #include <sys/types.h>
-#include "types.h"
 #include "librsync-config.h"
 
+#define HAVE_UINT64 1 /* assume success, but might be undefined below */
+
+#if defined(HAVE_STDINT_H)
+#  include <stdint.h>
```
dbaarda: I wonder if there is much point in supporting systems without stdint.h when blake2.h and rollsum.h both depend on it.

Is it possible for librsync.h and supporting libs to be compiled on a system with stdint.h and then deployed on a system without it? If so, then I guess it is worth it, but otherwise it makes things much simpler if we just require it.

src/librsync.h
```diff
@@ -37,9 +37,36 @@
...
+#else
+# if SIZEOF_UNSIGNED_INT == 4
```
dbaarda: Aren't all the defines that these depend on all defined in config.h? And isn't config.h in the same boat as types.h when it comes to what is deployed on systems?

I think librsync-config.h is deployed on systems and built by cmake specifically to avoid any dependencies on config.h.

sourcefrog: stdint is apparently from C99, and generally I think it's reasonable we rely on C99.  And, as you say, blake2.h assumes it.

I'll put up an alternative fix.

PR #92, opened 2016-10-23
Remove types.h, always use stdint.h

### LINKED ISSUE #90: types.h not found

sourcefrog:
Fixes #90


src/librsync.h
```diff
@@ -37,7 +37,7 @@
 #define _RSYNC_H
 
 #include <sys/types.h>
-#include "types.h"
+#include <stdint.h>
 #include "librsync-config.h"
```
dbaarda: Note that librsync-config.h includes only the typedef for rs_long_t and is built from librsync-config.h.cmake, using some  logic in CMakeLists.txt that is preceeded by a FIXME comment about how this is a bit ugly.

Now we have stdint.h we could also drop librsync-config.h and just define rs_long_t as intmax_t here.

There are arguments that off_t would be more appropriate (since it's a file offset, which is what off_t is for). However the standards are sufficiently fuzzy that there is also off64_t and loff_t, and which does what depends on things like defining "#define _FILE_OFFSET_BITS 64" or "#define _LARGEFILE64_SOURCE" and it's all compiler and platform dependent, so a mess. I think intmax_t is simple, encapsulates what we want, and should cause no problems.

PR #95, opened 2016-10-24
Faster Generic Hashtable.

dbaarda:
This is an alternative implementation to https://github.com/librsync/librsync/pull/79 that is about the same speed but maybe a little cleaner implementation with more tests. It also includes optional extra match stats with --stats, reduced memory usage by packing blocksigs using --sum-size, and much faster delta calculations for large files.

On an Intel(R) Atom(TM) CPU D525 @ 1.80GHz, for a 1GiByte file with 50% different data, using a 1KiByte block-size and 8Byte sum-size for a RELEASE build, it is about 4x as fast. Comparing the current master branch;

$ time ../rdiff -f -s -I 32768 -O 32768 delta /home/abo/tmp/sig.1024K /home/abo/tmp/new.1024K /home/abo/tmp/delta.1024K
librsync: loadsig statistics: signature[1048576 blocks, 1024 bytes per block] speed[12.6 MB (12.6 MB/s) in, 0.0 MB (0.0 MB/s) out, 1 sec]
librsync: delta statistics: literal[16384 cmds, 536870912 bytes, 49152 cmdbytes] copy[2 cmds, 536870912 bytes, 0 false, 18 cmdbytes] speed[1073.7 MB (1.8 MB/s) in, 536.9 MB (0.9 MB/s) out, 593 sec]
587.20user 3.74system 9:55.86elapsed 99%CPU (0avgtext+0avgdata 51396maxresident)k
2097168inputs+1048704outputs (1major+12528minor)pagefaults 0swaps

with this branch;

$ time ../rdiff -f -s -I 32768 -O 32768 delta /home/abo/tmp/sig.1024K /home/abo/tmp/new.1024K /home/abo/tmp/delta.1024K
librsync: loadsig statistics: signature[1048576 blocks, 1024 bytes per block] speed[12.6 MB (12.6 MB/s) in, 0.0 MB (0.0 MB/s) out, 1 sec]
librsync: match statistics: signature[537395200 searches, 524288 (0.098%) matches, 627469786 (1.168x) weak sum compares, 1548850 (0.288%) strong sum compares, 1547240 (0.288%) strong sum calcs]
librsync: delta statistics: literal[16384 cmds, 536870912 bytes, 49152 cmdbytes] copy[2 cmds, 536870912 bytes, 0 false, 18 cmdbytes] speed[1073.7 MB (7.6 MB/s) in, 536.9 MB (3.8 MB/s) out, 141 sec]
131.24user 3.78system 2:24.41elapsed 93%CPU (0avgtext+0avgdata 38552maxresident)k
2097168inputs+1048696outputs (1major+11378minor)pagefaults 0swaps

Note the extra match stats output, maxresident 38552K vs 51396K, and 141sec vs 593sec. For larger files it will be even faster.


src/hashtable.c
```diff
@@ -0,0 +1,163 @@
...
+
+    stats_inc(t->find_count);
+    do_probe(t, m, km) {
+        if (!(ke = t->ktable[i]))
+            return NULL;
+        stats_inc(t->keycmp_count);
+        if (km == ke) {
+            stats_inc(t->entrycmp_count);
+            if (!t->cmp(m, e = t->etable[i])) {
+                stats_inc(t->match_count);
```
MartinNowak: The fields should be renamed to hashcmp_count and cmp_count.

tests/largefile.test
```diff
@@ -0,0 +1,61 @@
...
+out="$datadir/out.$blocks"
+
+if [ ! -f "$old" ]; then
+   mkdir -p $datadir
+   dd bs=$blocks count=1024 if=/dev/urandom >"$old"
+   dd bs=$blocks count=256 if=/dev/urandom >"$new"
+   dd bs=$blocks count=256 skip=128 if="$old" >>"$new"
+   dd bs=$blocks count=256 if=/dev/urandom >>"$new"
+   dd bs=$blocks count=256 skip=640 if="$old" >>"$new"
+fi
```
MartinNowak: That test is a bit flawed, almost all data of A is contained in B. Real world cases might have as little as 50% in common (that was the figure for ubuntu-16.04-desktop-amd64.iso vs. ubuntu-16.04.1-desktop-amd64.iso).

tests/largefile.test
```diff
@@ -0,0 +1,61 @@
...
+
+run_test time $bindir/rdiff $debug -f -b 1024 -S 8 -s signature $old $sig
```
MartinNowak: With the number from below (1024K blocks at max), that's 1Mi entries at most, but 4MiB does fit in L3 cache (at least in mine), so you're not measuring the performance of your table when it moves to main memory.

tests/largefile.test
```diff
@@ -0,0 +1,61 @@
...
+run_test time $bindir/rdiff $debug -f -s -I 32768 -O 32768 delta $sig $new $delta
+run_test time $bindir/rdiff $debug -f -s patch $old $delta $out
```
MartinNowak: Please get accustomed to use long-options in scripts, greatly improves readability, short options are for the command line.

src/hashtable.c
```diff
@@ -0,0 +1,163 @@
...
+    h ^= h >> 13;
+    h *= 0xc2b2ae35;
+    h ^= h >> 16;
+    return h;
+}
+
+/* Get hash key, reserving zero as an empty bucket marker. */
+static inline unsigned get_key(const hashtable_t *t, const void *e)
+{
+    unsigned k = t->hash(e);
```
MartinNowak: Indirect function calls for hashing and comparison are a huge obstacle for the optimizer. Think of it, your calling a function to pull out the weak_sum field.
Not sure that separating the hashtable implementation (still somewhat specific) justifies such a performance problem.
Unfortunately this is crappy C and generic options are limited, also the strong comparison is configurable, but at least for the weak sum it seems unnecessary.

src/hashtable.h
```diff
@@ -0,0 +1,227 @@
...
+    cmp_f cmp;                  /* Function for comparing entries. */
+#ifndef HASHTABLE_NSTATS
+    /* The following are for accumulating hashtable_find() stats. */
+    long find_count;            /* The count of finds tried. */
+    long match_count;           /* The count of matches found. */
+    long keycmp_count;          /* The count of key compares done. */
+    long entrycmp_count;        /* The count of entry compares done. */
+#endif
+    void **etable;              /* Table of pointers to entries. */
+    unsigned ktable[];          /* Table of hash keys. */
```
dbaarda: This is true when you have a hash-value match, but it's not true when the hash values are different. If the hash values don't match, you don't need to look at the entry value, so you don't get an extra cache mise.

This is part of where I'm optimizing for lookup misses. For a 716M file with 50% different data, using 1K blocks, for a load factor of 0.7, we get the following match statistics;

librsync: match statistics: signature[375756800 searches, 366592 (0.098%) matches, 1053222017 (2.803x) weak sum compares, 867330 (0.231%) strong sum compares, 866681 (0.231%) strong sum calcs]

Note that there were 1,053,222,017 hash compares (2.803 per search) for 866,681 entry compares (0.231% per search). That's only  0.082% entry compares per hash compare, or one entry compare for every 1215 hash compares.

So by packing the hash values tighter, we get less cache misses for the hash compares, and there is way more of them than entry compares.

src/hashtable.c
```diff
@@ -0,0 +1,163 @@
...
+/* Open addressing works best if it can take advantage of memory caches using
+ * locality for probes of adjacent buckets on collisions. So we pack the keys
+ * tightly together in their own key table and avoid referencing the element
+ * table and elements as much as possible. Key value zero is reserved as a
+ * marker for an empty bucket to avoid checking for NULL in the element table.
+ * If we do get a hash value of zero, we -1 to wrap it around to 0xffff. */
+
+/* Use max 0.8 load factor to avoid bad open addressing performance. */
+#define HASHTABLE_LOADFACTOR_NUM 8
+#define HASHTABLE_LOADFACTOR_DEN 10
```
dbaarda: Yes, but with the tight packing of the hash values it doesn't hurt that much. For files with size/loadfactors of 512K/0.5, 716K/0.7, 819K/0.8 1024K/0.5 we get;

512K/0.5
=======
librsync: match statistics: signature[268697600 searches, 262144 (0.098%) matches, 313791392 (1.168x) weak sum compares, 518749 (0.193%) strong sum compares, 518434 (0.193%) strong sum calcs]
librsync: delta statistics: literal[8192 cmds, 268435456 bytes, 24576 cmdbytes] copy[2 cmds, 268435456 bytes, 0 false, 18 cmdbytes] speed[536.9 MB (7.5 MB/s) in, 268.5 MB (3.7 MB/s) out, 72 sec]

716K/0.7
=======
librsync: match statistics: signature[375756800 searches, 366592 (0.098%) matches, 1053222017 (2.803x) weak sum compares, 867330 (0.231%) strong sum compares, 866681 (0.231%) strong sum calcs]
librsync: delta statistics: literal[11456 cmds, 375390208 bytes, 34368 cmdbytes] copy[2 cmds, 375390208 bytes, 0 false, 18 cmdbytes] speed[750.8 MB (7.0 MB/s) in, 375.4 MB (3.5 MB/s) out, 108 sec]

819K/0.8
=======
librsync: match statistics: signature[429811200 searches, 419328 (0.098%) matches, 2103451089 (4.894x) weak sum compares, 1075564 (0.250%) strong sum compares, 1074655 (0.250%) strong sum calcs]
librsync: delta statistics: literal[13104 cmds, 429391872 bytes, 39312 cmdbytes] copy[2 cmds, 429391872 bytes, 0 false, 18 cmdbytes] speed[858.8 MB (6.2 MB/s) in, 429.4 MB (3.1 MB/s) out, 139 sec]

1024K/0.5
=======
librsync: match statistics: signature[537395200 searches, 524288 (0.098%) matches, 627469786 (1.168x) weak sum compares, 1548850 (0.288%) strong sum compares, 1547240 (0.288%) strong sum calcs]
librsync: delta statistics: literal[16384 cmds, 536870912 bytes, 49152 cmdbytes] copy[2 cmds, 536870912 bytes, 0 false, 18 cmdbytes] speed[1073.7 MB (7.5 MB/s) in, 536.9 MB (3.8 MB/s) out, 143 sec]

You can see that at loadfactor 0.8 it did get as high as averaging 4.894 hash compares per search and the execution time of 139secs was a fair bit worse than the 108secs the 716K/0.7 took, but the execution time was still better than the 143secs the 1024K/0.5 one took.

If I drop the max loadfactor to 0.7 the the match and delta stats become;

819K/0.4
=======
librsync: match statistics: signature[429811200 searches, 419328 (0.098%) matches, 327713444 (0.762x) weak sum compares, 1075564 (0.250%) strong sum compares, 1074655 (0.250%) strong sum calcs]
librsync: delta statistics: literal[13104 cmds, 429391872 bytes, 39312 cmdbytes] copy[2 cmds, 429391872 bytes, 0 false, 18 cmdbytes] speed[858.8 MB (8.3 MB/s) in, 429.4 MB (4.2 MB/s) out, 103 sec]

The execution time drops from 139secs to 103secs, which is a pretty good ~25% improvement, but not that astounding and at the cost of using significantly more memory. I'm happy to change it if people prefer the tradeoff for more speed though.

src/hashtable.c
dbaarda: I did a lot of dicking around with this and tried a heap of different things, some of which are visible in the commit history, and some of which are not. I tried a variety of things to avoid the function de-ref overheads and in the end was surprised at how little difference it made, though maybe that was early in the development before I'd optimized as much for the memory cache.

Trying it now with the following diff applied;

--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -81,7 +81,7 @@ static inline unsigned mix32(unsigned int h)
 /* Get hash key, reserving zero as an empty bucket marker. */
 static inline unsigned get_key(const hashtable_t *t, const void *e)
 {
-    unsigned k = t->hash(e);
+    unsigned k = *(unsigned *)e;
     return k ? k : -1;
 }

which removes the hash-func de-refencing and just assumes that the first field in the entry struct is the unsigned hash value (it is) we get the following performances;

### 512K/0.5 with no hash func

librsync: match statistics: signature[268697600 searches, 262144 (0.098%) matches, 313791392 (1.168x) weak sum compares, 518749 (0.193%) strong sum compares, 518434 (0.193%) strong sum calcs]
librsync: delta statistics: literal[8192 cmds, 268435456 bytes, 24576 cmdbytes] copy[2 cmds, 268435456 bytes, 0 false, 18 cmdbytes] speed[536.9 MB (8.3 MB/s) in, 268.5 MB (4.1 MB/s) out, 65 sec]

#### 716K/0.7

librsync: match statistics: signature[375756800 searches, 366592 (0.098%) matches, 1053222017 (2.803x) weak sum compares, 867330 (0.231%) strong sum compares, 866681 (0.231%) strong sum calcs]
librsync: delta statistics: literal[11456 cmds, 375390208 bytes, 34368 cmdbytes] copy[2 cmds, 375390208 bytes, 0 false, 18 cmdbytes] speed[750.8 MB (7.1 MB/s) in, 375.4 MB (3.5 MB/s) out, 106 sec]

Comparing this to the equivalent runs above, the execution time dropped for 512K/0.5 from 72secs to 65secs, and for 716K/0.7 from 108sec to 106sec. For low load factors it does make a small but measurable difference, but for higher load factors it's in the noise, and the execution time is completely dominated by the memory cache costs of doing on average 2.8 hash value comparisons per search.

I originally wanted to make the hashtable API as clean and generic as possible, which is why I went with the indirect function. I also at one point tried a different API where you provided the hash value to hashtable_add() and hashtable_find() and did the hashing externally. In the end I felt like the performance benefits were too small to bother.

src/hashtable.c
```diff
@@ -0,0 +1,163 @@
...
+    while ((i = (i + ++s) & mask) != index)
+
+void *hashtable_add(hashtable_t *t, void *e)
+{
+    assert(e != NULL);
+    do_probe(t, e, k) {
+        if (!t->ktable[i]) {
+            t->count++;
+            t->ktable[i] = k;
+            return t->etable[i] = e;
```
dbaarda: I did have exactly that early in the development but slowly evolved away from it, partly because of performance concerns that in retrospect may have been unfounded (I didn't realize how much memory cache locality mattered).

There was also some API related ugliness around it. Because the cmp() function is not a simple entry-to-entry compare, but a fancy deferred strongsum calculating comparison between a match object and an entry, it's not possible to use the same cmp() function to de-dupe entries in hashtable_add(). So the common hashtable_find_slot() routine had to use a different cmp() function.

src/hashtable.c
dbaarda: I found them very useful for getting a feel for what the search was doing and where the costs were. There is also support for turning them of by defining HASHTABLE_NSTATS, and all those stats_inc() calls get #defined to nothing (this also turns off the match stats support in sumset.c and rdiff).

I also found that there was no measurable difference when turning them off, so I've got them defaulting to on. We could add a CMake option to turn them off.

tests/largefile.test
dbaarda: I was copying the conventions used in the other tests. But yeah, you are right. I'll change them all to use long options, but maybe in another pull request.

tests/largefile.test
dbaarda: I'm using a pretty crappy Intel(R) Atom(TM) CPU D525   @ 1.80GHz which according to the following has only 1MiB of L2 cache;

http://ark.intel.com/products/49490/Intel-Atom-Processor-D525-1M-Cache-1_80-GHz

Certainly the execution times for the tests are already pretty painfully slow, and the performance is highly dependent on and dominated by any cache-localization changes I make.

Note that 1Mi entries ends up in a 0.5 loadfactor hashtable with 2Mi entries, which is 8MiB of ram for the hash value table, and 16MiB of ram for the entry pointer table (64bit architecture means 8bytes per pointer), which is much larger than my 1MiB L2 cache.

dbaarda:
A quick note for anyone looking at the timing of these tests in horror;

1) I'm intentionally using a crappy CPU to highlight bottlenecks. Modern workstation CPU's run these tests many times faster than this.

2) I'm using a 1K blocksize to maximize the stress on the hashtable. This makes it many times slower than using the recommended blocksize=sqrtt(filesize).

3) The files I'm reading/writing to are on nfs mounted from a QNAP NAS, though delta calculation is CPU and memory cache bound, so this actually has little affect.

dbaarda:
It's interesting you got slowdowns for smaller files. On my crappy Atom the new code was consistently faster than the old code even for small files. I suspect that on my platform even 32MiB files were stressing the L2 cache for the old hashtable, and the new hashtable has a bit better locality behavior. On a platform with more cache both old and new don't stress the cache and other factors start to dominate.

I'll have a look at some profiling and micro-optimizations to fix that after this goes in.

dbaarda:
Um... it appears I have permissions to merge this pull request, which surprises me. I thought only sourcefrog could do that.

Are people OK for me to merge this, or should I leave that up to sourcefrog?

dbaarda:
It's been a week since I asked if anyone would object to me merging this. If no one objects in the next day or so I'll merge it.

PR #96, opened 2016-11-07
Fix debian packaging scripts

Chilledheart:
I'm not sure this patch reflects all the changes made from 0.9.6-2. Any way it works for me.


sourcefrog:
Thanks, looks good.  We should probably fix the changelog entries including the author (you can use yourself).

Chilledheart:
Yeah, it seems it isn't perfect on the changelog entry which doesn't include 2 releases in the github release page. I'll update it the weekend (it's a bit late where I live, almost 3 am :-p).

dbaarda:
Just had a look at the changelog and the current Debian 0.9.7-10 changelog is this is missing quite a lot. There is also a 1.0.0-1~exp1 version in experimental created 09 Mar 2015 by Andrey Rahmatullin <wrar@debian.org> that probably has some more improvements/updates.

I think I'd like to pull in the current Debian changes first, and then apply updates to it. Ideally I'd like this to at least track Debian.

dbaarda:
I'll try to clean this up, possibly with some alternative pull requests, in the near future.

Chilledheart:
It will be very helpful to add the missing changelog. And I will rebase this pull request after that.

dbaarda:
@wRAR I've just started looking at updating the Debian package information and I'm a bit rusty on that stuff. I see you are the Debian package maintainer for librsync, would you be able to assist?

I was thinking I'd just grab the Debian directory from the 0.9.7 debian package sources, check that in as an update, and then refine it from there. However, there's stuff in there like patches that are no longer relevant and I have no idea of the latest-and-greatest Debian packaging practices.

I'm happy to take any advice or recommendations from you on this. The point is to try and get an updated release into Debian, so I'm happy to do whatever it takes to make that easier.  This is an important part of the final polish up before we can address #107 and make a new release.

wRAR:
@dbaarda Debian packaging happens separately from the upstream development and you are not required to include any packaging in the upstream releases.

dbaarda:
@wRAR so does having it in our github repo make it easier for you, or is it just annoying noise? Should we remove it, or is it actually useful in some way to someone?

wRAR:
@dbaarda no, it's not useful. Debian maintains the packaging files separately.

dbaarda:
I'm now kind of inclined to just delete the debian directory to resolve this. However, my one lingering doubt is perhaps keeping it and updating it makes it easier for people to build their own deb's from github sources between Debian releases. I'd like to hear peoples opinions before I do anything.

I'm also inclined to think that resolving this should not block a new release, given that distro maintainers are going to ignore our debian stuff anyway.

dbaarda:
@wRAR  In the interest of closing out old pull requests I'm taking another look at this.

I understand that Debian packagers have their own separate packaging files that are maintained separately, but I still feel that including our own minimal Debian packaging stuff so that people can easily build their own customized debs would be useful, particularly when librsync releases include fixes and features that are not yet available in the official Debian releases.

However, I'm not entirely sure what minimal Debian packaging stuff would look like... would it look like this patch? I've had a look at your v2.0.2 packaging and it's quite different  (and I suspect more up to date) than what's in this pull request. I've never been 100% familiar with Debian packaging, and what I did know is almost certainly out of date.

Should I just apply this pull request, and then update it further from there? Or would it be best to just abandon this and do the deb-helper setup again from scratch? Or should I just copy your v2.0.2 Debian configs? Any hints would be most welcome.

PR #98, opened 2017-02-21
Optimize and refine hashtable.

dbaarda:
This shaves about 10~15% off the delta execution time, winning back the tiny performance losses we got for small files when we switched to the new open addressing hashtable.

The commit history shows the incremental refinements made. Interestingly the change from function pointers to directly referenced functions made nearly zero difference, which surprised me. The big wins were in making things static inline, with an early small win for changing the probing to a for loop that doesn't check for looping through all elements, instead relying on hashtable_add() to always leave one empty bucket.

It now is effectively a template/parametric class implemented using preprocessor macros, which is pretty much the only way to make it possible to static inline all the hashtable_find() related calls while keeping it generic enough for wider use.

I also made the comments all doxygen compatible. The doxygen docs can be previewed here;

http://minkirri.apana.org.au/~abo/librsync/hashtable_8h.html

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -427,7 +427,7 @@ EXTRACT_PACKAGE        = NO
 # included in the documentation.
 # The default value is: NO.
 
-EXTRACT_STATIC         = NO
+EXTRACT_STATIC         = YES
 
 # If the EXTRACT_LOCAL_CLASSES tag is set to YES, classes (and structs) defined
 # locally in source files will be included in the documentation. If set to NO,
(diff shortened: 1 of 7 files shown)


dbaarda:
FWIW I think I figured out why the change from function pointers to directly referenced functions didn't make any difference to the performance. Those functions were not marked static (and they could have been, as the equivalents were in the old implementation), which means the compiler optimization can't decide to inline them, since they need to be callable from external compilation units. A static function is only callable from within that compilation unit, so the optimizer is free to munge it inline if it decides that is optimal.

The final solution in this pull request marks them static inline. The inline bit might be redundant for a good optimizing compiler in this case, but probably helps for less-smart compilers.

dbaarda:
With the last two changes applied that makes some functions static, the speed improvement is closer to 15~20%. This more than wins back the 9~12% regression for small files from the new hashtable.

dbaarda:
Ping? It looks like I have perms to  merge this without further comments by anyone. 

If no one objects in the next week I'll assume people are happy with this and will merge it.

MartinNowak:
Thanks for getting back to this @dbaarda.

PR #100, opened 2017-05-01
Windows - Linux build fixes for 2.0.1 version

marius-nicolae:
Hi,

These are some small fixes for successful building on Windows, with MSVC 2015, and Linux 64, using gcc 5.4.0.

What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,6 +250,13 @@ add_dependencies(check ${LAST_TARGET} isprefix_test rollsum_test hashtable_test 
 
 enable_testing()
 
+if (NOT MSVC)
+	# It appears the "_fileno" is defined only on Windos (Microsft Visual Studio). In the rest of the platforms it need
+	# to be "fileno". For this, the _fileno variable is set here and a corresponding "#cmakedefine" directive is inserted
+	# in config.h.cmake.
+	set(_fileno fileno)
+endif (NOT MSVC)
+
 # Create conf files
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/config.h.cmake ${CMAKE_CURRENT_BINARY_DIR}/src/config.h)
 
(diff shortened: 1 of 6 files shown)


dbaarda:
Ugh... C libraries and non-posix systems.

It seems fileno() is not C99 but is posix, and _fileno() is ISO C++. Given that this is being used together with the non c99 posix fstat() and librsync is C not C++, I think the code should use fileno() and have a #define that renames fileno() to _fileno() on the platforms that don't have fileno().

The #define should be based on a HAVE_FILENO check in CMakeLists.txt, not just a platform check.

dbaarda:
In #101  I've discussed how we should fix all the filesystem platform dependencies by moving them all into fileutil.c. In #108 I've just merged in the pointer arithmetic fixes. That leaves the def file as the last thing in this patch to consider.

I'm not familiar with Visual Studio and how important this def file is, but I guess I could be easily convinced. If this merge request was reduced to just the def file I'd merge it.

marius-nicolae:
If the def file is not desired then all the exported functions can be declared with MSVC specific __declspec(dllexport/import) keywords. Concretely, for MSVC, in the header file a macro should be defined like:
--8<
 #if _DLL
    #define DLLEXPORT __declspec(dllexport)
#else
    #define DLLEXPORT __declspec(dllimport)
#endif
>8--

In the header file, all the exported symbols need to be prepended with the  DLLEXPORT macro and while building the dll the "/D _DLL" macro need to be added in the cl.exe command line.

dbaarda:
I think I'd prefer a def file than cluttering the code with MSVC specific `#ifdefs`.

dbaarda:
Everything in this pull request except for the def file is now included in #111, along with a few other file compatibility stuff and tidyups.

Since this still includes something worth keeping I'm not going to close this now, but it will need to be reduced to just the def file changes after #111 is merged.

dbaarda:
In #156 a new LIBRSYNC_EXPORT definition was added and used on everything exported by librsync.h to explicitly export the public symbols. This definition is windows aware and defines things equivalent to the DLLEXPORT idea described above when _WIN32 is defined.

I believe this means that the def file is no longer required. That was the last feature in this pull request that was not already implemented, so I'm now closing this pull request.

PR #73, opened 2016-05-03
Add support for _fseeki64 (MSC).

gulikoza:
Add support for _fseeki64 (MSVC, mingw64). Should fix some problems in #69 as well.

Let's see if this works in AppVeyor :-)

PR #101, opened 2017-05-25
Use fseeko64 when compiling with Mingw

charlievieth:
This PR defines `fseek` as `fseeko64` if it is available when compiling with Mingw - fixing an integer overflow induced error on Windows when working with large files.

The issue is that with Mingw the size of `off_t` is 4 (both when compiling for 32 and 64 bit) and both `fseeko` and `fseeko64` are defined - causing `fseek` to be defined as `fseeko`.  This leads to an overflow when the `rs_long_t` offset in `rs_file_copy_cb` is passed to `fseek`.

I noticed that the [librsync 2.0.1](https://github.com/librsync/librsync/blob/master/NEWS.md#librsync-201) section of NEWS.md says that `fseeko64` will be used for Mingw in for the 2.0.1 release.  I was wondering if we could get this change or the one planned for 2.0.1 into the master branch before then?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,7 +32,7 @@ NOT RELEASED YET
 
  * `popt` is only needed when `rdiff` is being built. (gulikoza)
 
- * Use `fseeko64` on mingw. (gulikoza)
+ * `fseeko64` is used when compiling with mingw.
 
  * `rdiff -s` option now shows bytes read/written and speed. (gulikoza)
 
--- a/src/buf.c
+++ b/src/buf.c
@@ -53,19 +53,13 @@
 #include "job.h"
 #include "util.h"
 
-/* Use fseeko instead of fseek for long file support if we have it.
- *
- * If provided use fseeko64 when compiling with mingw64 on Windows
- * as the size of off_t is 4 and will overflow on large files.
+/* Use fseeko64 or fseeko instead of fseek for long file support if
+ * we have it.
  */
-#if defined(__MINGW32__) && defined(HAVE_FSEEKO64)
-    #define fseek fseeko64
-#else
-    #ifdef HAVE_FSEEKO
-    #define fseek fseeko
-    #elif defined HAVE_FSEEKO64
-    #define fseek fseeko64
-    #endif
+#if defined(HAVE_FSEEKO64) && defined(HAVE_FOPEN64)
+#  define fseek fseeko64
+#elif defined(HAVE_FSEEKO)
+#  define fseek fseeko
 #endif
 
 /**
--- a/src/buf.c
+++ b/src/buf.c
@@ -53,11 +53,19 @@
 #include "job.h"
 #include "util.h"
 
-/* use fseeko instead of fseek for long file support if we have it */
-#ifdef HAVE_FSEEKO
-#define fseek fseeko
-#elif defined HAVE_FSEEKO64
-#define fseek fseeko64
+/* Use fseeko instead of fseek for long file support if we have it.
+ *
+ * If provided use fseeko64 when compiling with mingw64 on Windows
+ * as the size of off_t is 4 and will overflow on large files.
+ */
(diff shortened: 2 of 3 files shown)


dbaarda:
I've just had a look at this and am wondering if it wouldn't be better to either just prefer fseeko64 when fseeko also exists, or do something checking SIZEOF_OFF_T vs SIZEOF_OFF64_T to decide what to use.

IMHO it's nearly always better to choose things like this based on the actual constraints that dictate the choice than special-casing the platform. Platforms can fix their non-standard behaviour, and/or there might be other platforms that have this non-standard behaviour.

charlievieth:
@dbaarda Thanks for the quick response and I entirely agree with your point about actual vs. platform constraints. 

I considered giving fseeko64 preference over fseeko, but the [glibc documentation for fseeko64](https://www.gnu.org/software/libc/manual/html_node/File-Positioning.html) gave me pause (emphasis mine):

> This function is similar to fseeko with the only difference that the offset parameter is of type off64_t. This also **requires that the stream stream was opened using either fopen64, freopen64, or tmpfile64** since otherwise the underlying file operations to position the file pointer beyond the 2^31 bytes limit might fail.

I'll test giving fseeko64 preference over fseeko using glibc, but won't have any time to do it until early next week.  If replacing fseeko with fseeko64 works without required files to be opened with fopen64 and friends this is probably the easiest solution.  

Otherwise, checking for `SIZEOF_OFF_T < SIZEOF_OFF64_T` and exposing those macros via the cmake config would work just as well.

dbaarda:
It sounds like if we use fseeko64 we also need to use fopen64. Perhaps something like this;
```c
#if defined(HAVE_FSEEKO64) && defined(HAVE_FOPEN64)
#  define fopen fopen64
#  define fseek fseeko64
#elif defined(HAVE_FSEEKO)
#  define fseek fseeko
#endif
```

charlievieth:
@dbaarda The define specified in your last post works perfectly.  I tested it in on Darwin, Windows Server 2016 (MingGW-64) and Ubuntu 16.04 32bit (GCC 5.4.1 and GLIBC 2.23) - just to make sure there weren't any weird interactions on the 32bit system - and it works.  

I updated the commit to include your change and to define the `fopen` as `fopen64` in [fileutil.c](https://github.com/librsync/librsync/blob/master/src/fileutil.c#L73-L87).

dbaarda:
I cant see any checks or defines for HAVE_FOPEN64 in CMakeLists.txt or config.h.cmake. I think they will need to be added for this to really work correctly. Alternatively we assume that fopen64() exists if fseeko64 does.

It looks like we may also need support for _fseeki64() as pull request #73 was trying to add.

There is also related looking ifdefs for using fstat() vs fstat64() vs _fstati64() in util.c.

It feels bad to me to reproduce all this ifdef logic in every *.c file that uses these (fileutil.c uses fopen, buf.c uses fseek, util.c uses fstat). Is this something we could put on the end of config.h.cmake, or is that bad practice that might break importing of the corresponding system's stdio.h etc headers? (note we always import "config.h" before any system <*.h> headers). Perhaps we should put it into fileutil.h (and make buf.c import it)?

While we are at it, perhaps  rs_get_filesize() doesn't belong in util.[ch] but in fileutil.[ch].

dbaarda:
Oh, found another one in patch request #100, fileno() vs _fileno() in buf.c and rs_get_filesize() in util.c.

If we could bundle all these file related platform ifdefs into fileutil.h and move rs_get_filesize() into fileutil.[ch], only buf.c would need to include fileutil.h to get the right fseek() and fileno().

charlievieth:
Sounds good, and apologies for the delay - I'll update the PR.

dbaarda:
I've now put together #111  which includes everything in this Patch request and several other file compatibility patches, along with some other tidy ups.

I'm going to close this patch request in favor of #111.

Issue #77, opened 2016-06-26
Compile error rs_stats


hojjatjafary:
Hi,
I wrote a simple code that only calls "rs_sig_file", but I get this compile error:
error C3646: 'start': unknown override specifier
the 'start' is a member of "rs_stats" inside the "librsync.h", but when I include <time.h> or <ctime> above the "librsync.h" include then the project compiles successfully.
Why I need to include time library inside my own cpp file that uses only a function call of "rs_sig_file", I also tried to include "stdlib.h" that is works too.
I compile using Visual studio 15.

```
#include <stdlib.h>
//#include <ctime>
#include "rsync/librsync.h"
#include "sync_signature.hpp"

int SyncSignature::generateSignature(std::string inputFilePath, std::string sigFilePath)
{
    FILE *fpInput;
    fpInput = fopen(inputFilePath.c_str(), "rb");

    FILE *fpSig;
    fpSig = fopen(sigFilePath.c_str(), "wb+");

    // The MD4 sig magic is deprecated.
    rs_result res = rs_sig_file(fpInput, fpSig, 2048, 32, RS_BLAKE2_SIG_MAGIC, NULL);

    fclose(fpInput);
    fclose(fpSig);

    return 0;
}
```

sourcefrog:
Yes it seems like `librsync.h` ought to include `time.h`

dbaarda:
I've included this fix in pull request #104. I'll merge it if I don't get any complaints in a day or so.

dbaarda:
This should be fixed now. Please reopen if you find otherwise.

PR #104, opened 2017-06-01
Use C99 stdint.h and inttypes.h.

dbaarda:
This removes librsync-config.h which is no longer needed, since rs_long_t just uses the C99 intmax_t type.

I audited and tidied up all the format strings, using FMT_* definitions in trace.h that used C99 standard PRIxNN definitions for format strings. This also tidied up all the rs_log() calls.

I also found some questionable uses of int that should have been using size_t that I fixed.

There was one potential bug in src/tube.c for large files with large copies on platforms where rs_long_t is larger than size_t that I also fixed.

My main concern with this Patch request is using "%zu" for formatting size_t with FMT_SIZE. This is a C99 thing that might not be supported on MS compilers. We may need to add some #ifdef platform checks in trace.h where this is defined.


dbaarda:
I did some research and found out "MS Visual Studio didn't support %zu printf specifier before VS2013. Starting from VS2013 (e.g. _MSC_VER >= 1800) %zu is finally available." However, according to the following [MSDN page](https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx) Visual Studio 2015 does not support %zu.

So it looks like I need to add some CMakeLists.txt magic for this.

dbaarda:
cmake magic added... if noone objects in the next couple of days I'll merge this.

dbaarda:
Added a tiny fix for issue #77

Issue #6, opened 2014-07-23
Speed issues: Use larger buffer sizes.


pavel-odintsov:
Hello!

Could you optimize rdiff tool performance? It's very slow and unusable in most cases =(

paulharris:
This speed issue may be resolved with latest master, due to merge of improvements related to large files.

dbaarda:
This bug was complaining about signature generation time, which is pretty simple file-read and checksum calculation work. It's unsurprising that setting buffer size larger than block size more than doubled his throughput.

That the CPU gets pegged doing md4sum calculations is unsurprising. I don't think much can be done to optimize the md4sum calculation. The best that can be done is parallelize the md4sum calculation so we can peg multiple CPU's :-) Also maybe the new blake2sums' are faster... dunno.

At the end he mentions delta calculation, which has been significantly improved with a new hashtable for large files since this bug was filed, though using 1M blocksize would have avoided problems with the old hashtable.

The idea of parallelizing signature generation is being tracked in #68.

The only thing left that this bug could do is have the buffer size default to something like 16x block size, not just 16K, if unspecified.

dbaarda:
I've been thinking a little about the bufferlen settings and am about to do something about this.

The rs_inbuflen and rs_outbuflen vars are publicly visible/setable according to librsync.h. They primarily control the  rs_buffers_t size used by the whole-file API implemented in whole.c, but crucially rs_outbuflen  is also used by delta.c as the maximum size buffered/output per output-literal command.

In scoop.c librsync keeps it's own internal scoop_buf buffer that data is copied into for matching against (the double-copying this implies is another problem). In delta.c we always copy all the data in the input buffers to the scoop. This buffer grows by doubling size till it is large enough to contain a full block for matching against + the largest amount of miss data buffered for a single output-literal command + the full size of the next input buffer. This means it grows as large as the next power of 2 larger than rs_outbuflen + block_size+rs_inbuflen (note we could just pre-allocate the max needed size).

For rs_inbuflen it's only used for pre-fetching data to be copied into the scoop. After a match we always need at least a whole block of data to begin checking for matches again, so having rs_inbuflen < block_size is inefficient. 

For rs_outbuflen The maximum output the algorithm generates in a single iteration is the largest output-literal command (cmd bytes+literal data).

[...]

PR #105, opened 2017-06-15
Improve default buffer sizes used.

dbaarda:
In delta.c make the max miss output length 4*block_len  instead of rs_outbuflen so that it's independent of the whole file API buffer length arguments.

In whole.[ch] make rs_inbuflen and rs_outbuflen default to zero which means use recommended buffer sizes for the operations being performed. Make all the whole file operations use recommended buffer sizes for the operation being performed unless the rs_inbuflen or rs_outbuflen overrides have been set.

This makes rdiff use more efficient buffer sizes by default when --input-size and --output-size arguments are not set.

### DIFF
--- a/src/delta.c
+++ b/src/delta.c
@@ -306,14 +306,15 @@ static inline rs_result rs_appendmatch(rs_job_t *job, rs_long_t match_pos, size_
  * Append a miss of length miss_len to the delta, extending a previous miss
  * if possible, or flushing any previous match.
  *
- * This also breaks misses up into rs_outbuflen segments to avoid accumulating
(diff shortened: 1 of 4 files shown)


dbaarda:
Note this is the final fix needed to close issue #6.

I'd be interested to know how people find the performance with these changes, particularly when people are playing with different (mostly large) blocksizes.

dbaarda:
I would like to see some more work on figuring out optimal buffer sizes for the different operations, but perhaps they can be refined in followup pull requests. This sets the framework up for tuning it later.

If no-one objects in the next couple of days I'm going to merge this.

dbaarda:
OK... no objections... I'm merging this now.

PR #106, opened 2017-06-19
#include <stdio.h> in librsync.h

santazhang:
Include what you use: librsync.h uses FILE struct, but did not #include <stdio.h> to import the symbol.


dbaarda:
Thanks. This is pretty straightforward, so merging now.

PR #108, opened 2017-07-05
Fixes for building as a subdirectory using MSVC compiler

lasalvavida:
Changes ${CMAKE_SOURCE_DIR} to ${CMAKE_CURRENT_SOURCE_DIR} so that it is relative to the project, not a top-level calling cmake.

Corrects some compilation issues with MSVC as the compiler. You cannot increment on a `void *` pointer (I'm pretty sure that's invalid C, so I'm not sure why gcc accepts it). Also defines `S_ISREG` on windows systems.


lasalvavida:
Just saw #100. Noting that there is some overlap here.

dbaarda:
This patch is broken because casting pointers to int isn't safe on some platforms (64bit platforms). I believe #100 does this more correctly by casting to (char*) for doing the pointer arithmetic. I must admit I didn't quite understand why that was required at the time I looked at it, but your description of this patch helped me understand why it was required.

For your S_ISREG define bit, I have a comment on #101 where I think we need to move all the filesystem related platform dependent stuff into fileutil.c and centralise it there. The guy doing #101 (charlievieth) agreed and said he'd try to get to it. I think I'd like to see all of this done in one patch, either #101 or another if someone else gets to it first.

Your CMAKE_CURRENT_SOURCE_DIR change makes perfect sense and we should do that.

TLDR; I want all the file handling fixed in a single patch that moves it all to fileutil.c as discussed in #101. If you fix the pointer arithmetic to use casts to (char*) instead of (int) as is done in #100 or even if you drop it to just the CMAKE_CURRENT_SOURCE_DIR, I'll accept this.

lasalvavida:
@dbaarda, I've made the requested changes

dbaarda:
Thanks for that. This is now pretty simple and is passing all tests so I'm merging it now.

PR #109, opened 2017-07-20
Add public rs_sig_args() function for recommend signature args from filesize.

dbaarda:
This adds an rs_sig_args() function to librsync.h that can be used to get recommended block_len and strong_len arguments to use for a signature from the old filesize. This is then used by the whole file operations and rdiff for the default --block-size and --sum-size arguments.

This uses the same algorithm used by rsync for the default block_len and and strong_len but is more conservative to reduce the risk of hash collisions (rsync has a whole-file checksum and multiple retries with different hash seeds to mitigate that, librsync doesn't). 

This makes rdiff produce significantly smaller signature files and calculate much faster deltas by default while ensuring they are still safe against random hash collisions. However, it also makes it more vulnerable against hostile files crafted to produce hash collisions that would result in those files being corrupted.

There are a few minor tidy-ups I could make to this pull request before submitting it, but I'm posting it now in the hope of prompting discussion about "should rdiff default to safe+fast signature settings, or to paranoid safe settings", and if the proposed API is OK.

An alternative API would be to provide an alternative rs_sig_begin() method (rs_sig_begin2? rs_sig_begin_fs?) that also takes the old filesize to calculate default block_len and/or strong_len arguments if they are zero. Note this API approach could also be used to provide an alternative rs_sig_load() method that takes the signature file size for pre-allocating the blocksum storage instead of the currently non-public approach of setting the filesize in the job struct.


dbaarda:
Because this is potentially contentious I'm not going to merge this before the next release unless people explicitly request it.

dbaarda:
The current version of this pull request has the following rdiff behaviour;

* `-S 0` means "use the maximum full length of the hash used".
* `-S <n>` argument specifies a minimum strongsum size, and a longer strongsum will be used if the filesize recommends it.
* A minimum recommended strongsum size of 12 bytes will be used if the filesize is unknown (eg a pipe to stdin).
* The minimum recommended strongsum size for a small file (<1 block) is 5 bytes.

In practice this means it's not possible to set a strongsum size <5 for any file, or a strongsum size <12 for piped input. This might be a problem.

People might really want to explicitly set the strongsum size to a small value, either because they know the piped input is small, or they really do want to risk hash collisions.

Do we really want rdiff to treat the `-S <n>` argument as a minimum hash size? Or should we always use it directly and do something like `-S -<n>` to mean "recommended, at least <n>"?

Also the minimum recommended strongsum of 5 bytes comes from assuming new_filesize=old_filesize+16MB and using +2bytes safety margin. The +16MB behaves like a miniumum +3bytes. If we dropped that the minimum would be 2bytes (for a 1byte filesize). With RabinKarp the number of useful bits in the weaksum is much improved... do we need the min strongsum length to be 5bytes?

dbaarda:
I'm not happy with this yet. I don't particularly like the changed meaning of rdiff's `-S` strongsum length argument. I'm also not 100% with the librsync API for this.

I'm happy with the behaviour of the `-B` blocksize argument though, and I think that's were most of the wins with no risks lie.

I guess I need to think on this a bit more... feedback/advise are welcome.

dbaarda:
I've changed it to use `-S=-1` to mean use minimum strong_len for the given file size and block_len. This reverts to the master branch behavior of not treating `-S` as a minimum, but using it directly. This means it is still possible to set strong_len's small enough to be unsafe, but I've added a rs_log() warning if you do this.

I think this is ready to merge. The rdiff behavior is unchanged from the master branch except that `-1` can be used as the strong_len to mean "use minimum", and you get warnings logged if you set the strong_len small enough to be unsafe. It possibly also includes a bugfix for rdiff around getopt parsing of int's into size_t that would bite people on non-little-endian platforms.

I'm going to merge this after I update NEWS.md

dbaarda:
For some reason the travis test results are not currently showing on github, but I can see them on travis-ci.org and they are passing clean.

I'm going to merge this now.

ericzolf:
Hi, we use rdiff to check that our usage of the librsync is correct but in the recent past has started to change its behaviour, see https://github.com/rdiff-backup/rdiff-backup/issues/304. Despite a typo, fixing the rollsum-algorithm was easy enough, but I'm at loss to guess the sum-size value which would give us the old behaviour. Can you help me out?

dbaarda:
@ericzolf I'm really sorry this caused you guys problems... replying on that bug for you.

Always feel free to file a bug asking for assistance...I will generally respond pretty quickly.

PR #112, opened 2017-08-28
Don't add duplicate blocks to the hashtable.

dbaarda:
A large number of duplicate blocks (eg long runs of zero) act like a DOS attack on the open addressing hashtable because they all cluster.

This means we only keep the first copy of duplicate blocks in the hashtable, and thus matches in the delta will always refer to the first matching block.

This does mean we can't return a single long match for a long sequence of identical blocks, and instead will return multiple single block matches against the first block. However, that's what we currently do anyway since we only find the first matching block in the hashtable even if there are multiple identical blocks in there.

It's also debatable if a single long match of multiple identical blocks is more efficient than multiple small matches against the first block. The delta might be a little smaller with a long match, but applying the delta gets better disk-cache hits for multiple references to the first block.

### DIFF
--- a/src/hashtable.h
+++ b/src/hashtable.h
@@ -234,6 +234,18 @@ static inline void _FUNC(_free)(hashtable_t *t)
     _hashtable_free(t);
 }
 
+/** Initialize hashtable stats counters.
+ *
+ * This will reset all the stats counters for the hashtable,
+ *
+ * \param *t - The hashtable to initializ stats for. */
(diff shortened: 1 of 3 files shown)


dbaarda:
This is pretty strait-forward, is a performance improvement, and resolves remaining problems in #4.

If no one objects in a couple of days I'm going to merge this.

dbaarda:
OK... no complaints so I'm merging.

PR #114, opened 2017-09-08
Add support for using libb2 blake2 implementation.

dbaarda:
This adds support for finding and optionally using libb2.

Adds a USE_LIBB2 option if libb2 is found to enable using it. It defaults to OFF because currently the reference implementation seems to be significantly faster.


dbaarda:
This is pretty simple and by default is a no-op since it defaults to using the included reference blake2 implementation even if libb2 is found.

If no one objects in the next couple of days I'm going to merge this.

dbaarda:
No objections, one thumbs up! merging now.

PR #116, opened 2017-09-26
CMakeLists.txt: Don't require a C++ compiler

bkuhls:
By default, CMake assumes that the project is using both C and C++.  By
explicitly passing 'C' as argument of the project() macro, we tell CMake
that only C is used, which prevents CMake from erroring out if a C++
compiler doesn't exist.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@
 # Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
 
-project(librsync)
+project(librsync C)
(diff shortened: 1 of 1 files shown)


bkuhls:
This patch fixes errors detected by buildroot autobuilders, like
http://autobuild.buildroot.net/results/3ca/3cacb87e39a0ebb9d227fff4be71bea8132bf0ed/
http://autobuild.buildroot.net/results/3ca/3cacb87e39a0ebb9d227fff4be71bea8132bf0ed/build-end.log

dbaarda:
This seems fine to me. I'm going to merge it now.

Issue #115, opened 2017-09-24
Unable to create delta due to sig hashtable assertion error


jaknb:
## Problem
I'm using librsync 2.0.1 and I'm facing the following error trying to create a delta:
```
rs_delta_begin: Assertion `!(sig)->hashtable || (sig)->hashtable->count == (sig)->count' failed.
```

## How to recreate
**PLEASE NOTE**: This will create a 1GB dummy file
``` bash
fallocate -l 1G big.dat
rdiff signature big.dat sigfile
rdiff delta sigfile big.dat deltafile
```
*Now you should encounter the problem.*

Clean the dummy files:
``` bash
rm big.dat sigfile deltafile
```

## System info (Ubuntu 16.04)
```
computer                  
    description: Computer
    product: VirtualBox
    vendor: innotek GmbH
    version: 1.2
    serial: [REMOVED]
    width: 64 bits
    capabilities: smbios-2.5 dmi-2.5 vsyscall32
    configuration: family=Virtual Machine uuid=[REMOVED]
  *-firmware
       description: BIOS
       vendor: innotek GmbH
       physical id: 0
       version: VirtualBox
       date: 12/01/2006
       size: 128KiB
       capabilities: isa pci cdboot bootselect int9keyboard int10video acpi
  *-memory
       description: System memory
       physical id: 1
       size: 3949MiB
  *-cpu
       product: Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz
       vendor: Intel Corp.
       physical id: 2
       bus info: cpu@0
       width: 64 bits
       capabilities: fpu fpu_exception wp vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx rdtscp x86-64 constant_tsc rep_good nopl xtopology nonstop_tsc pni pclmulqdq ssse3 cx16 sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx rdrand hypervisor lahf_lm abm 3dnowprefetch rdseed clflushopt
  *-disk
       description: ATA Disk
       product: VBOX HARDDISK
       physical id: 0.0.0
       bus info: scsi@2:0.0.0
       logical name: /dev/sda
       version: 1.0
       serial: [REMOVED]
       size: 16GiB (17GB)
       capabilities: partitioned partitioned:dos
       configuration: ansiversion=5 logicalsectorsize=512 sectorsize=512 signature=f14c5211
```

[...]

PR #117, opened 2017-09-29
Fix broken signature assert when there are duplicate blocks.

dbaarda:
This fixes an assert statement broken by pull request #112 which doesn't add duplicate blocks to the hashtable, which now means the hashtable can have less entries than the number of blocks in a signature.

This adds a test file for the existing changes.test to check this, and fixes the assert  statement in rs_signature_check().


dbaarda:
This fixes #115  which breaks debug builds.

Since this is kinda important and trivial in size, I'm going to merge this tomorrow if no one objects before then.

PR #119, opened 2017-10-10
Fix #32 hang for truncated input.

### LINKED ISSUE #32: Infinite loop for truncated signature or delta files.
This might be related to #29 
I was investigating why delta files seem to have additional characters on the end when generating deltas of identical files...

With the current rdiff,

``` bash
echo "Hello world" > input
rdiff signature input sig
rdiff delta sig input delt
rdiff patch input delt output
# all is fine
# time to hang the process, truncate the delta file
dd if=delt of=delt2 bs=1 count=1
rdiff patch input delt2 output
# HANG, infinite loop somewhere
# Note that when the delta file is big enough, THEN the problem is handled properly
dd if=delt of=delt2 bs=1 count=4
rdiff patch input delt2 output
librsync: ERROR: (librsync) patch job failed: unexpected end of input
librsync: ERROR: unexpected end of input
```

dbaarda:
Fix scoop.c and mksum.c handling of truncated input to correctly terminate with RS_INPUT_ENDED. This means rdiff will correctly return an error message instead of hanging for any type of operation that hits an unexpected EOF.

Add small optimizations to scoop.c and tube.c input handling.

Remove input and output size arguments for rdiff delta operation from largefile.test to test the now sane default values.

Update NEWS.md with this and all other missing changes ready for the v2.0.1 release.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,21 +5,18 @@
 NOT RELEASED YET
 
  * Extensively reworked Doxygen documentation, now available at
-   http://librsync.sourcefrog.net/
-   (Martin Pool)
+   http://librsync.sourcefrog.net/ (Martin Pool)
 
  * Removed some declarations from librsync.h that were unimplemented or no
-   longer ever useful: `rs_work_options`, `rs_accum_value`.
-   Remove declaration of unimplemented `rs_mdfour_file()`.
-   (Martin Pool)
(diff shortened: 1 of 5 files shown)


dbaarda:
No one screamed, this fixes a bug, merging it now.

PR #121, opened 2017-10-11
Fix #13 use slack deltas for zero sized signatures.

### LINKED ISSUE #13: rdiff inefficient for zero sized signatures.
Hello!

I tried to use rdiff with zero sized signatures (with correct header but without any literals/copy items) and rdiff delta for this signatures produce bunch of cpu consumption and need so much time.

dbaarda:
This uses slack deltas for signatures of empty files. This means it doesn't bother doing rollsum calculations and hashtable lookups against an empty hashtable, instead just emitting literal commands. This makes it heaps more efficient. I added an empty input file for the changes.test to test this case.

I also removed a vestigial tests/dot file (leftover from autoconf?), and updated testcommon.py to include '0' in the tested buffer sizes so that it tests the recommended default buffer sizes, which depend on the block size and type of operation being performed.


dbaarda:
This is a tiny fix for a bug. Merging now.

PR #122, opened 2017-10-11
Fix #33 rs_job_iter needs calling twice with eof=1.

### LINKED ISSUE #33: rs_job_iter needs to be called twice with eof=1

dbaarda:
This fixes #33 and tidies up and optimizes the rs_job_iter() code.  It removes code checking for programming errors and replaces them with more extensive precondition asserts.

I also slightly cleaned and optimized rs_tube_catchup() when there is nothing to catchup.

It also includes changing rs_job_check() from a function to a #define so it gets completely optimized away when debugging is turned off and more clearly indicates where the job was bad when asserts fail.


dbaarda:
This is another small bug fix. Not much point in waiting for it, I'm going to merge it now.

PR #123, opened 2017-10-12
Fix cmake/FindPOPT.cmake to mark_as_advanced POPT_INCLUDE_DIRS

dbaarda:
Previously it was incorrectly marking the non-existent POPT_INCLUDES as advanced.

This ensures that ccmake doesn't list POPT_INCLUDE_DIRS as a configurable option.


dbaarda:
one-liner... merging.

PR #126, opened 2017-10-15
Fix #55 remove excessive rs_fatal() calls.

### LINKED ISSUE #55: rs_fatal is overused
librsync has a function/macro `rs_fatal` etc, to indicate severe errors or internal bugs, which calls `abort()`.

However as a well-behaved library, it's probably better if it doesn't do that unless there's no safe way to cleanly unwind.  Probably most uses should be pulled out and it should just return an error.

dbaarda:
Replace unnecessary fs_fatal() calls in emit.c with asserts.
These errors could only happen if rs_int_len() was broken, so don't
need runtime checks, asserts are sufficient.

This also involved slighly changing the logic to avoid compiler
warnings when assserts are omitted for release builds, which also made
the code a little more consistent.

The only rs_fatal() calls left are for alloc failures, the tube "popping", netint.c encountering a >64bit int file offset or size, and delta.c failing a paranoid weaksum check with the rdiff --paranoia checking.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -82,6 +82,12 @@ NOT RELEASED YET
  * Fixed #33 so rs_job_iter() doesn't need calling twice with eof=1.
    Also tidied and optimized it a bit. (dbaarda,
    https://github.com/librsync/librsync/issues/33)
+   
+ * Fixed #55 remove excessive rs_fatal() calls, replacing checks for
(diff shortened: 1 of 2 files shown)


dbaarda:
This is a small bugfix with NEWS entry for v2.0.1 ready. Merging now to ensure it goes in before next release.

PR #127, opened 2017-10-16
Tune max miss size for deltas.

dbaarda:
Change the max miss size to 32K from 4*block_len.

This gives a constant 0.01% 3 command bytes overhead for misses. For small block sizes it has nearly no impact on performance but gives slightly smaller deltas. For large block sizes it gives a tiny performance increase and significantly reduced memory requirements for deltas.

Also add extra block size argument to largefile.test to make it easy to test the impact for different block sizes.

This had less performance impact than I expected, but it is a little win.


dbaarda:
This is a tiny change with a small but measurable performance, memory, and delta size improvement.

No NEWS entry is required as this falls under the already documented collection of "improve performance" changes. Merging now so it goes in before next release.

PR #133, opened 2017-10-30
Use the GNUInstallDirs CMake module to get the install paths.

wRAR:


CMakeLists.txt
```diff
@@ -124,11 +125,11 @@ if(APPLE)
   set(CMAKE_MACOSX_RPATH ON)
   set(CMAKE_SKIP_BUILD_RPATH FALSE)
   set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
   set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-  list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
+  list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" isSystemDir)
   if("${isSystemDir}" STREQUAL "-1")
```
dbaarda: What exactly does this if section do? It looks to me like it sets CMAKE_INSTALL_RPATH to exactly the same thing it was set to above.

wRAR: Yes, I'm wondering that too. The relevant commit is 75f6884884a9a8b2561eacf6ff7839c15bed77a6.

wRAR: Actually, that code is directly from https://cmake.org/Wiki/CMake_RPATH_handling#Always_full_RPATH

dbaarda: Ah, OK. It still doesn't make sense to me, but sure.

dbaarda:
I'm going to merge this as a minimal change to use the proper cmake install module stuff.

Further work that would be nice is adding the install for the man pages which is missing. However, that's something for another pull request.

PR #136, opened 2017-12-23
Avoid check_c_source_runs() when cross compiling

fornwall:
Using check_c_source_runs() causes the build to fail when cross compiling. In that case we cannot check for printf "%zu" size_t formating support by executing a binary - instead assume it.

This should be safe as AFAIK only MS Visual Studio versions prior to VS2013 doesn't support it, and it's unlikely that someone should be using an old Visual Studio to cross compile librsync.


dbaarda:
I'm not sure if there is a better way to do this than just assume it works for cross-compiling, but I'll accept it as better than what we currently have.

PR #143, opened 2018-02-26
Fix #47 Updated RPM spec file for 2.0.2

### LINKED ISSUE #47: rpm spec file is out of date

deajan:
Updated spec file for current master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -395,4 +395,11 @@ install(FILES
 
 message (STATUS "CMAKE_C_FLAGS  = ${CMAKE_C_FLAGS}")
 
+install(FILES
+	doc/librsync.3
+	DESTINATION ${CMAKE_INSTALL_MANDIR}/man3)
+install(FILES
+	doc/rdiff.1
+	DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
+
 # vim: shiftwidth=4 expandtab
--- a/librsync.spec
+++ b/librsync.spec
@@ -36,7 +36,7 @@ This package contains header files necessary for developing programs
 based on librsync.
 
 %prep
-#wget --no-check-certificate --timeout=5 -O %{_sourcedir}/%{name}-master.zip %{gitsource}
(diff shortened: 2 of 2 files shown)


deajan:
I actually also have a file for 2.0.1 with CMakeLists.txt path variable fixes. Is master "as good as" 2.0.1 so there's no need to make an extra PR for 2.0.1 ?

dbaarda:
I'm about to make a 2.0.2 release so don't bother with 2.0.1.

What are the path fixes you mention? I think the CMakeLists.txt is missing standard man page install stuff that would be nice to fix.

deajan:
The path fixes were supposed to be for 2.0.1. That was before I've seen that you already fixed them in current master. I had modded CMakeLists.txt in order to replace hard coded 'lib' with ${INSTALL_LIB_DIR}, 

Indeed, there's no man page install stuff, so I actually had to add 4 ugly lines to the spec file in order to add the manpages in the rpm.

dbaarda:
Would it be better to fix the CMakeLists.txt to correctly install the man pages instead of the ugly fix to the spec file? We already include(GNUInstallDirs) so I think it's as simple as adding this towards the end;

install(FILES doc/librsync.3
    DESTINATION ${CMAKE_INSTALL_MANDIR}/man3)
install(FILES doc/rdiff.1
    DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)

dbaarda:
I'll make this the last pull I merge before releasing v2.0.2. As soon as you've added the CMakeLists.txt changes I suggest above and remove the hack from the spec file, I'll merge this and do the release.

deajan:
I've added the CMakeLists.txt changes, removed the spec file hack and updated the version schema for the upcomming 2.0.2 release. SPEC file builds successfully, inluding the man files :)

deajan:
I've seen PR #142 and removed perl as build dependancy too.

dbaarda:
Thanks. I'm merging this now, and will then put together the 2.0.2 release.

PR #147, opened 2018-06-06
Fix memory leak on signature->block_sigs

telles-simbiose:
I noticed that there was a memory leak on rs_job_iter, after investigating I found that it was not freeing memory hold by `rs_signature_t->block_sigs`. 

This patch fixed the memory leak.

### DIFF
--- a/src/sumset.c
+++ b/src/sumset.c
@@ -168,6 +168,8 @@ void rs_signature_done(rs_signature_t *sig)
 {
     hashtable_free(sig->hashtable);
     rs_bzero(sig, sizeof(*sig));
+
+    free(sig->block_sigs);
 }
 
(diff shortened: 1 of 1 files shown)


src/sumset.c
```diff
@@ -168,6 +168,8 @@ void rs_signature_done(rs_signature_t *sig)
 {
     hashtable_free(sig->hashtable);
     rs_bzero(sig, sizeof(*sig));
+
+    free(sig->block_sigs);
```
dbaarda: You need to do this before the sig is zero'ed... so put it after hashtable_free() and before rs_bzero().

telles-simbiose:
Done @dbaarda .
It was wrong because I did it in my application and then just moved to sumset.c

PR #153, opened 2019-05-22
Document delta file format

zmj:


dbaarda:
Thanks for doing this.

PR #156, opened 2019-08-16
Fix #130 don't export private symbols.

### LINKED ISSUE #130: Please don't export private symbols
It seems there are a lot of symbols which are private and not a part of the public ABI that are exported. Those include `_hashtable_*`, `blake2b_` (when the bundled b2 is used), and probably some of `rs_*` functions (at least some of `rs_*` functions were exported in 1.0.0 but are not exported anymore because they were made `static` but SONAME was not changed so I assume they were private from the beginning).
Setting all symbols as hidden and only explictitly marking public ones as visible helps maintaining ABI correctly and understanding when the SONAME must be bumped.

dbaarda:
This uses cmake's GenerateExportHeader support to generate a src/librsync_export.h and uses it to explicitly export all public symbols. It also sets the "C_VISIBILITY_PRESET hidden" target property for librsync to hide all targets that are not explicitly exported. The install now needs and includes the src/librsync_export.h which is imported by librsync.h.

This required updating rdiff.c to only use publicly exported symbols, and adding some things to librsync.h for public export that rdiff.c needs. In the process of doing this, the error output from rdiff was tidied up to be much nicer.

It also involved removing rs_licence_string from librsync.h and docs, since it was declared as an extern, but not actually defined anywhere.


dbaarda:
Ugh, it seems the travis checker uses old versions of cmake with https://gitlab.kitware.com/cmake/cmake/issues/15829 that breaks GenerateExportHeaders. Lets see if the newer xenial build environment fixes it.

dbaarda:
travis shows xenial fixes it for linux builds, but osx builds are still broken. There is no newer osx environment we can use.

This means we either ignore the osx breakage, or we do something that works with older cmake versions, like not use GenerateExportHeaders.

I await wRAR's comments/advise as a distribution packager on this.

dbaarda:
It's a sad world when you tools for isolating you from the differences between platforms is causing problems with different platforms :-(

I think I'm going to have to abandon GenerateExportHeaders and use a hand-coded librsync_export.h instead.

wRAR:
Unfortunately, I don't have experience with `GenerateExportHeader`. If you don't want to require a new enough cmake then it seems that writing the header manually is the only way.

As for packaging, I can only list the symbols used in Debian by binaries linked to librsync:

```
rs_build_hash_table
rs_delta_begin
rs_delta_file
rs_file_copy_cb
rs_free_sumset
rs_job_drive
rs_job_free
rs_job_iter
rs_loadsig_begin
rs_loadsig_file
rs_patch_begin
rs_patch_file
rs_sig_begin
rs_sig_file
```

Looks like all of them are marked as exported in this PR.

I also think the soname should be bumped, even though only "private" symbols were removed, but I'm not 100% sure about the accepted practices.

dbaarda:
wRAR: thanks for that info, particularly the list of what people are using which is interesting.

I've switched to using a manual librsync_export.h and bumped the minor version from 2.0.2 to 2.1.0 to reflect the additional functions added to librsync.h. I note that in our doc/versioning.md we say we use semantic versioning but only for the interface defined in librsync.h. After this change is in we should be correctly start semantic versioning only the exported symbols.

If there is a technical packaging reason why this change should bump the major version to v3.0.0 I'm open to arguments, but for now I'm assuming only bumping the minor version in accordance with our documented practices is enough.

wRAR:
The only reason to bump the soname in this case is "maybe some people have some binaries that have used the private symbols".

dbaarda:
wRAR: Thanks again for your prompt reply. If anyone has been using "private symbols" then they've been doing it wrong anyway, so I thint 2.1.0 it is.

I'm going to merge this now, and will be doing a 2.1 0 release soon.

PR #161, opened 2019-08-19
fix build on MinGW-w64 and MSVC

Adsun701:
This sets rsync_EXPORTS on Windows to prevent build failure. In addition, missing LIBRSYNC_EXPORT types are added to certain variables in librsync.h to prevent redefiniton errors on MSVC, and dlls are correctly installed into the bin directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,7 @@ add_test(NAME hashtable_test COMMAND hashtable_test)
 add_executable(sumset_test
     tests/sumset_test.c src/sumset.c src/util.c src/trace.c src/hex.c
     src/checksum.c src/rollsum.c src/mdfour.c src/hashtable.c ${blake2_SRCS})
+target_compile_options(sumset_test PRIVATE -DLIBRSYNC_STATIC_DEFINE)
 target_link_libraries(sumset_test ${blake2_LIBS})
 add_test(NAME sumset_test COMMAND sumset_test)
 
@@ -361,7 +362,11 @@ set_target_properties(rsync PROPERTIES
     VERSION ${LIBRSYNC_VERSION}
     SOVERSION ${LIBRSYNC_MAJOR_VERSION}
     C_VISIBILITY_PRESET hidden)
-install(TARGETS rsync ${INSTALL_TARGETS_DEFAULT_ARGS} DESTINATION ${CMAKE_INSTALL_LIBDIR})
+install(TARGETS rsync ${INSTALL_TARGETS_DEFAULT_ARGS}
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+)
 
 
 ########### next target ###############
@@ -378,7 +383,11 @@ if (BUILD_RDIFF)
     message (WARNING "Popt library is required for rdiff target")
   endif (POPT_FOUND)
 
-  install(TARGETS rdiff ${INSTALL_TARGETS_DEFAULT_ARGS} DESTINATION ${CMAKE_INSTALL_BINDIR})
+  install(TARGETS rdiff ${INSTALL_TARGETS_DEFAULT_ARGS}
+          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+          LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+          ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  )
 endif (BUILD_RDIFF)
 
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@
 
 NOT RELEASED YET
 
+ * Fix MSVC builds by adding missing LIBRSYNC_EXPORT to variables in
+   librsync.h, add -DLIBRSYNC_STATIC_DEFINE to the sumset_test target,
+   and correctly install .dll files in the bin directory.
+   (adsun701, https://github.com/librsync/librsync/pull/161)
+
 ## librsync 2.1.0
 
 Released 2019-08-19
--- a/src/librsync_export.h
+++ b/src/librsync_export.h
@@ -1,6 +1,9 @@
 #ifndef LIBRSYNC_EXPORT_H
 #  define LIBRSYNC_EXPORT_H
 
+#ifdef LIBRSYNC_STATIC_DEFINE
+#  define LIBRSYNC_EXPORT
+#else
 #  ifdef _WIN32
 #    ifdef rsync_EXPORTS
 #      define LIBRSYNC_EXPORT __declspec(dllexport)
@@ -10,5 +13,6 @@
 #  else
 #    define LIBRSYNC_EXPORT __attribute__((visibility("default")))
 #  endif
+#endif
 
 #endif                          /* LIBRSYNC_EXPORT_H */
--- a/src/rdiff.c
+++ b/src/rdiff.c
@@ -77,24 +77,6 @@ enum {
 
 char *rs_hash_name;
 
-const struct poptOption opts[] = {
-    {"verbose", 'v', POPT_ARG_NONE, 0, 'v'},
-    {"version", 'V', POPT_ARG_NONE, 0, 'V'},
-    {"input-size", 'I', POPT_ARG_INT, &rs_inbuflen},
-    {"output-size", 'O', POPT_ARG_INT, &rs_outbuflen},
-    {"hash", 'H', POPT_ARG_STRING, &rs_hash_name},
-    {"help", '?', POPT_ARG_NONE, 0, 'h'},
-    {0, 'h', POPT_ARG_NONE, 0, 'h'},
-    {"block-size", 'b', POPT_ARG_INT, &block_len},
-    {"sum-size", 'S', POPT_ARG_INT, &strong_len},
-    {"statistics", 's', POPT_ARG_NONE, &show_stats},
-    {"stats", 0, POPT_ARG_NONE, &show_stats},
-    {"gzip", 'z', POPT_ARG_NONE, 0, OPT_GZIP},
-    {"bzip2", 'i', POPT_ARG_NONE, 0, OPT_BZIP2},
-    {"force", 'f', POPT_ARG_NONE, &file_force},
-    {0}
-};
-
 static void rdiff_usage(const char *error, ...)
 {
     va_list va;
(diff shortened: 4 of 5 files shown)


CMakeLists.txt
```diff
@@ -145,6 +145,11 @@ if (CMAKE_C_COMPILER_ID MATCHES "(Clang|Gnu|GNU)")
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
 endif()
 
+# set rsync_EXPORTS on WIN32
+if(WIN32)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Drsync_EXPORTS")
+endif()
```
dbaarda: I was under the impression cmake defines this automatically;

https://stackoverflow.com/questions/27429732/cmake-adds-dlibname-exports-compile-definition

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 NOT RELEASED YET
 
+ * Fix generic Windows builds by setting rsync_EXPORTS in CMakeLists.txt,
```
dbaarda: This is actually adding LIBRSYNC_EXPORT to all exported variables in librsync.h, not just "certain variables". So this would be a better description (assuming librsync_EXPORT definition is not required);

* Fix MSVC builds by adding missing LIBRSYNC_EXPORT to variables in librsync.h, and correctly install .dll files in the bin directory.

CMakeLists.txt
Adsun701: You're right, it does; however the build failures are due to the different targets such as `sumset_test` and `rdiff` not having `rsyncs_EXPORTS` defined for them.

NEWS.md
Adsun701: Fixed.

CMakeLists.txt
```diff
@@ -260,6 +260,7 @@ add_test(NAME hashtable_test COMMAND hashtable_test)
 add_executable(sumset_test
     tests/sumset_test.c src/sumset.c src/util.c src/trace.c src/hex.c
     src/checksum.c src/rollsum.c src/mdfour.c src/hashtable.c ${blake2_SRCS})
+target_compile_options(sumset_test PRIVATE -Drsync_EXPORTS)
```
dbaarda: Ugh! This is because we are statically linking the test, instead of dynamically linking it against the librsync library.

Currently librsync_export.h and librsync.h assumes we only build a librsync shared library and dynamically link against that.

We can't just change sumset_test to dynamically link against the librsync library because it tests functions/methods/data that is not publicly exported by librsync.

So I think we need to change librsync_export.h to support the static linking case. As I understand it, when statically linking the correct thing is to have LIBRSYNC_EXPORT defined as empty.

I'll have a look at best practices and see what I can find.

CMakeLists.txt
```diff
@@ -377,8 +382,13 @@ if (BUILD_RDIFF)
   else (POPT_FOUND)
     message (WARNING "Popt library is required for rdiff target")
   endif (POPT_FOUND)
+  target_compile_options(rdiff PRIVATE -Drsync_EXPORTS)
```
dbaarda: Are you sure this is needed? As I understand it, when dynamically linking against the librsync.dll you need to NOT define `rsync_EXPORTS` so that it correctly tags the declarations with `__declspec(dllimport)`.

Are you forcing a static build perhaps? If you are, I think this would also be fixed by making librsync_exports.h support static builds.

CMakeLists.txt
dbaarda: It looks like cmake's GenerateExportHeader (which we can't use because it's broken for C on all but the most recent cmake versions) adds this at the beginning of its librsync_export.h:

```C
#  ifdef LIBRSYNC_STATIC_DEFINE
#    define LIBRSYNC_EXPORT
#    define LIBRSYNC_NO_EXPORT
#  else
```

And then you are expected to define `LIBRSYNC_STATIC_DEFINE` for targets that are statically linked. I think we should copy that... can you include the change to librsync_export.h in this pull request?

Adsun701: Added change to `librsync_export.h`, and added `-DLIBRSYNC_STATIC_DEFINE` to `sumset_test` target.

CMakeLists.txt
Adsun701: Added `-DLIBRSYNC_STATIC_DEFINE` to `rdiff` target.

src/librsync_export.h
```diff
@@ -1,6 +1,10 @@
 #ifndef LIBRSYNC_EXPORT_H
 #  define LIBRSYNC_EXPORT_H
 
+#ifdef LIBRSYNC_STATIC_DEFINE
+#  define LIBRSYNC_EXPORT
+#  define LIBRSYNC_NO_EXPORT
```
dbaarda: Actually, we never use LIBRSYNC_NO_EXPORT, so you can delete this line.

This macro is created by GenerateExportHeader for cases when you explicitly want to not export something and have configured your default as "public". We set our default to "hidden" and only use LIBRSYNC_EXPORT.

CMakeLists.txt
```diff
@@ -377,8 +382,13 @@ if (BUILD_RDIFF)
   else (POPT_FOUND)
     message (WARNING "Popt library is required for rdiff target")
   endif (POPT_FOUND)
+  target_compile_options(rdiff PRIVATE -DLIBRSYNC_STATIC_DEFINE)
```
dbaarda: We don't want rdiff statically linked, we want it to dynamically link against the rsync.dll or librsync.so library.

So we don't want this line. However, I'm curious to know if you found you needed this for some reason under MinGW and/or MSVC. If you do need it for some reason, I suspect there is another problem we need to fix for MinGW and/or MSVC on win32 platforms.

NEWS.md
```diff
@@ -4,6 +4,11 @@
 
 NOT RELEASED YET
 
+ * Fix MSVC builds by adding missing LIBRSYNC_EXPORT to variables in
+   librsync.h, add -DLIBRSYNC_STATIC_DEFINE to the sumset_test and
+   rdiff targets, and correctly install .dll files in the bin directory.
+   (adsun701, https://github.com/librsync/librsync/pull/161)
```
dbaarda: Remove the bit about rdiff here... we want rdiff to dynamically link against the rsync library.

We only required sumset_test to be statically linked because it explicitly tests functions that are not exported by the rsync library.

Adsun701: Fixed.

CMakeLists.txt
Adsun701: Fixed. Without the static define, rdiff failed to compile first due to "initializer element is not constant" for `rs_inbuflen` and `rs_outbuflen` (which are defined in `whole.c`), and then when I added second definitions for them in `rdiff.c`, a `function is marked dllimport` error resulted due to the LIBRSYNC_EXPORT qualifier. Ultimately, I fixed this by removing the `rs_*` prefixes and the qualifier from both variables in rdiff.c, making them distinct from the variables declared in `librsync.h` and defined in `whole.c`.

src/librsync_export.h
Adsun701: Fixed.

src/rdiff.c
```diff
@@ -62,6 +62,9 @@
 #include "librsync.h"
 #include "isprefix.h"
 
+/** Whole file IO buffer sizes, from whole.c */
```
dbaarda: Err... we don't want to do this. The whole point of setting the rs_inbuflen and rs_outbuflen defined in whole.c and exported from the rsync dynamically linked library in librsync.h is those are the variables used by whole.c for the input and output buffer sizes.

If we define and set our own inbuflen and outbuflen here in rdiff.c, it will have no effect on the buffer sizes used by whole.c in the librsync library.

I wondered if this was a limitation of DLL's... perhaps it's not possible to export variables from a dll, only functions? However a quick search seems to indicate that it is possible;

https://stackoverflow.com/questions/3097548/exporting-global-variables-from-dll

There must be something else going wrong if this is not working.

dbaarda: Actually, this raises questions about rs_librsync_version, which is currently declared in librsync.h, but is missing the LIBRSYNC_EXPORT qualifier there, and is defined in version.c with LIBRSYNC_EXPORT, and is used in rdiff.c.

How come you are not getting errors related to rs_librsync_version? I would expect that we need to add LIBRSYNC_EXPORT to it in librsync.h too.

CMakeLists.txt
dbaarda: Sorry, I didn't see this comment until just now... it was hidden as "resolved". Defining new variables in rdiff.c is definitely wrong, but I'm not sure why you hit the problems you did.

That error message is strange, and doesn't seem to correspond to what the online searches for that error say it's about. I have a sneaking suspicion that the compiler is getting confused because we have 2 variables exported/defined on the same line with a single LIBRSYNC_EXPORT for them both.

Perhaps edit whole.c AND librsync.h and put them each on their own line with their own LIBRSYNC_EXPORT, and see if that works.

Another thing that might be worth trying is adding "static" to their definition, though that shouldn't be needed.

Adsun701: Doesn't work. The `dllimport` attribute may actually [be the reason](http://mingw.5.n7.nabble.com/error-with-a-function-declared-with-dllimport-and-used-in-a-struct-td13576.html) for the `initializer element is not constant` error.

src/rdiff.c
Adsun701: I can confirm that `rs_inbuflen` and `rs_outbuflen` cannot be initialized in `rdiff.c` due to [limitations](https://docs.microsoft.com/en-us/cpp/c-language/rules-and-limitations-for-dllimport-dllexport) of the `dllimport` attribute.

Adsun701: I've decided to mark this as "[WIP]" until I can fix the `rs_inbuflen`/`rs_outbuflen` `initializer element is not constant` issue, because everything else seems to be resolved.

CMakeLists.txt
dbaarda: I found similar docs about "initializer element is not constant error.", but the explanations didn't seem to match the situation we have here; rs_inbuflen and rs_outbuflen are not constants, and they are being initialized to constant '0' values in whole.c.

However, I think I now understand why this is happening. The problem is not strictly rs_inbuflen or rs_outbuflen, it's "const struct poptOption opts[]" in rdiff.c that we are initializing with a value that includes '&rs_inbuflen" and "&rs_outbuflen" which are unknown at compile time, since it's only after linking that their addresses are known. This appears to be a limitation of dll's, and linux's dynamic linking doesn't seem to have this limitation.

To fix this is a bit tricky. I think probably the neatest way is to move the whole opts definition inside main() in rdiff.c, so that it is initialized at runtime. Other options include declaring local int's for inbuflen and outbuflen to put into opts (like you did before) and make rdiff_options() set rs_inbufflen = inbuflen and rs_outbuflen=outbuflen after all the arguments are parsed to make sure the cmdline arguments are injected into whole.c where they will be used (which you didn't have before). I think making opts be initialized at runtime is neater though, and we should perhaps add a comment explaining why.

I've also figured out why rs_librsync_version was not causing problems without LIBRSYNC_EXPORT defined on it in librsync.h; it's because version.c where it is defined doesn't import librsync.h, and thus the compiler doesn't see the miss-matched declaration/definition when it's compiling version.c. However, for the others whole.c and sumset.c DO import librsync.h, so it's important that the declarations in librsync.h match the definitions in the *,c files.

The main thing that doesn't make sense to me so far is that it appeared to work with LIBRSYNC_STATIC_DEFINE set for compiling rdiff. Everything I've read suggests that symbols imported from DLLs need to have __declspec(dllimport) defined on them to work. For it to appear to work without it suggests that it was somehow statically linking rdiff, but the way the cmake stuff is setup I can't see how it would have been doing that.

To be strictly correct, rs_librsync_version should have LIBRSYNC_EXPORT defined on it in librsync.h too. I see you've done this... thanks.

I think we've finally figured out how to make this work.

Adsun701: Fixed. Thanks for the help!

PR #164, opened 2019-09-11
Fix #120 use single-byte literal commands for small inserts.

### LINKED ISSUE #120: Use single-byte delta literal commands for small literals.
I noticed that although prototab.c defines a heap of single byte literal commands we are never using them, instead always using RS_OP_LITERAL_N1 instead, which uses 2 bytes. This means small litterals with 64 or less bytes are using a 2 byte command instead of a single byte command.

We are both not emitting single byte literal commands, and also not correctly interpreting them. This means adding support for this will produce deltas that are not backwards compatible with older rdiff versions. Unfortunately the current handling of this will try to interpret the following literal data as an integer length of literal data, producing at best a logged error message and assert failure (asserts are disabled in release builds) before proceeding with corrupt data, maybe halting with RS_CORRUPT later, but possibly completing with corrupt results.

So adding support for this may require a bump in the delta magic to warn about this.

dbaarda:
This means we now emit the one-byte RS_OP_LITERAL_[1-64] commands for short literal inserts in deltas. Note delta files that use these are backwards compatible with older versions of librsync/rdiff, as patch has always supported them.

Includes a changes.test input file that exercises this.


dbaarda:
This is pretty simple, passes all checks, and fixes an issue. I'm going to merge it now.

PR #165, opened 2019-09-11
remove warnings in mdfour.c, checksum.h, trace.h

Adsun701:
This removes warnings for `mdfour.c` (using unsigned long long to avoid casting warning on Windows due to different data models), `checksum.h` (give a return value of 0 after the switch statement to avoid `control reaches end of non-void function`), and `trace.h` (set `HAVE_PRINTF_Z` to off when cross-compiling for Windows in `CMakeLists.txt`)


src/mdfour.c
```diff
@@ -228,7 +228,7 @@ inline static void rs_mdfour_block(rs_mdfour_t *md, void const *p)
    buffer first. */
 inline static void rs_mdfour_block(rs_mdfour_t *md, void const *p)
 {
-    unsigned long ptrval = (unsigned long)p;
+    unsigned long long ptrval = (unsigned long long)p;
```
dbaarda: Casting to a long long can be a little expensive on platforms where addresses are smaller than that. The point of this was to figure out if the pointer p is aligned for a uint32_t. Copying it into another variable to do this is unnecessarily complicated. After reading the following;

https://stackoverflow.com/questions/15868313/why-cant-you-do-bitwise-operations-on-pointer-in-c-and-is-there-a-way-around-t
https://stackoverflow.com/questions/1845482/what-is-uintptr-t-data-type

I think a better answer is to remove this variable entirely and change the 'if' statement that uses it below to;

if ((uintptr_t)p & 3) {

My only concern is uintptr_t is "optional" in C99 (which is what we target as our minimum compiler requirement), and I have no idea if there are critical platforms that don't have it. If there are platforms without it, I suspect we should work around that with a mix of cmake and config.h.

CMakeLists.txt
```diff
@@ -110,8 +110,13 @@ check_type_size ( "unsigned short" SIZEOF_UNSIGNED_SHORT )
 
 # Check for printf "%zu" size_t formatting support.
 if(CMAKE_CROSSCOMPILING)
-  set(HAVE_PRINTF_Z ON)
-  message (STATUS "Cross compiling - assuming printf \"%zu\" size_t formatting support")
+  if(WIN32)
```
dbaarda: Is there a reason why the 'check_c_source_runs(...)' check below can't automatically handle this? I suspect that cmake can't do "runs" tests when cross-compliling... is that correct?

src/checksum.h
```diff
@@ -130,6 +130,7 @@ static inline rs_weak_sum_t weaksum_digest(weaksum_t *sum)
     case RS_RABINKARP:
         return rabinkarp_digest(&sum->rk);
     }
+    return 0;
```
dbaarda: I wonder if these fragments would be faster in the following form;
```C
  if (sum->kind == RS_ROLLSUM)) 
    return mix32(RollsumDigest(&sum->rs));
  return rabinkarp_digest(&sum->rk);
```
The main reason why I didn't code them all that way is it doesn't extend as nicely if we ever add another rollsum type. However, I suspect we will never add another rollsum... RabinKarp is pretty close to optimal on collisions and clustering.

Salamek: Or use `default:` if you want to keep `switch` in code...

PR #166, opened 2019-09-16
disable HAVE_PRINTF_Z when directly compiling on Windows

Adsun701:
This fixes warnings when directly compiling on Windows rather than just cross-compiling. The CheckCSourceRuns check for HAVE_PRINTF_Z somehow results in a success for HAVE_PRINTF_Z when directly configuring on Windows, resulting in multiple format warnings during compilation.


dbaarda:
Thanks again for this. It's pretty strange. Are you sure that the HAVE_PRINTF_Z checks weren't being cached from an earlier (perhaps cross-compiled) cmake run? Try purging your git client and trying again with;

$ git clean -fdx
$ cmake .

Assuming it wasn't as simple as that, I've done some tidyups and added comments, and I'll probably merge this soon.

Adsun701:
@dbaarda I am pretty sure HAVE_PRINTF_Z checks were not being cached from an earlier run. Thanks!

PR #168, opened 2019-10-02
Use -1 for "unknown" for file sizes.

dbaarda:
Change rs_file_size() to return -1 for unknown file sizes and update docstring in librsync.h. In sumset.c update rs_signature_init() work correctly for sig_fsize<12.

Update sumset_test.c to use -1 for unknown filesizes and add some tests for checking invalid blake2 strong_len arguments.


dbaarda:
This cherrypicks a change from #109 that is independent of that change, and really should have been like this from the start.

Because it's pretty trivial I'm going to merge it now.

Issue #154, opened 2019-07-31
How to build using mingw32 on Linux?


Hacksawfred3232:
Requesting this because I need rdiff as part of a requirement on a python lib I'm building. The API that my lib talks to requires a rsync signature file along with the main file or a rsync delta when uploading to their services. I only have a small server with about 3 GB of memory and a old processor, so I can't use WINE. Don't even know if cygwin works under wine and if anything that is built under cygwin can be ran on plain Windows.

mosolovsa:
I'm not sure if this is what you want, but:

> cd ~/librsync-2.0.2

(directory with CMakeFile.txt located
>mkdir build
>cd build
>cmake -DMINGW=ON -DCMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY -DCMAKE_C_COMPILER=i686-w64-mingw32.static-gcc ..

`-DCMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY` - to prevent compiling simple executable and trying to run it during cmake build.
i686-w64-mingw32.static-gcc must be path to your mingw c compiler.

>make rsync

Hope it helps...

dbaarda:
With the merge of #161 a bunch of problems building with MinGW have been fixed. It should now be possible to build it using mingw32 and cmake without forcing a static library.

Please try it and let us know if there are any remaining problems or tricks to make it work. I welcome any pull requests with documentation to make this easier.

Hacksawfred3232:
Sorry for not responding sooner! Trying to compile with mosolovsa's command minus the "STATIC_LIBRARY" flag returns this:
```
CMake Error at /usr/share/cmake-3.10/Modules/CMakeTestCCompiler.cmake:52 (message):
  The C compiler

    "/usr/bin/x86_64-w64-mingw32-gcc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /opt/buildservice/SOURCE/librsync/CMakeFiles/CMakeTmp

    Run Build Command:"/usr/bin/make" "cmTC_1285f/fast"
    /usr/bin/make -f CMakeFiles/cmTC_1285f.dir/build.make CMakeFiles/cmTC_1285f.dir/build
    make[1]: Entering directory '/opt/buildservice/SOURCE/librsync/CMakeFiles/CMakeTmp'
    Building C object CMakeFiles/cmTC_1285f.dir/testCCompiler.c.o
[...]

PR #169, opened 2019-10-02
Add cmake BUILD_SHARED_LIBS option for static library support.

dbaarda:
BUILD_SHARED_LIBS defaults to ON, and can be set to OFF using `ccmake .` to
build librsync as a static library.


dbaarda:
I've also updated the .travis.yml test builds to use more updated conventions/settings and added a test for static builds.

dbaarda:
This is related to issue #154 but I'm not sure that it fixes it.

PR #170, opened 2019-10-02
fix compilation with MS Visual Studio ad ignore build files

ardovm:
The Visual Studio C++ compiler seems not to like the current implementation of the `_KEY_HASH` macro. This pull request is aimed to restore compatibility with Microsoft Visual Studio 2019.

In addition, some wildcards are added to `.gitignore` targeting output files from CMake and from the build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,8 @@ CMakeDoxygenDefaults.cmake
 
 # Test binaries
 *_test
+*_test.exe
+*_test.exe.manifest
 
 *.bak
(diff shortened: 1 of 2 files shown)


dbaarda:
I did some digging and it seems "compound statement expressions" are a gcc extension that is not part of the c99 standard.

So this was my bad, and thanks for fixing it.

I'm going to merge now.

ardovm:
That was quick! Thank you!

PR #181, opened 2019-11-23
CMakeLists: fix build with old compilers

texierp:
The librsync code uses some C99 constructs, but old compilers such
as gcc 4.7 don't default to -std=c99. This commit makes sure librsync
is built with -std=gnu99.

Fixes:

src/rabinkarp.h:82:5: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode
     for (size_t i = len; i; i--) {
     ^
src/rabinkarp.h:82:5: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,7 @@ if (CMAKE_C_COMPILER_ID MATCHES "(Clang|Gnu|GNU)")
   # TODO: Set for MSVC and other compilers.
   # TODO: Set -Werror when the build is clean.
 
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -std=c99")
 endif()
 
 site_name(BUILD_HOSTNAME)


CMakeLists.txt
```diff
@@ -149,7 +149,7 @@ if (CMAKE_C_COMPILER_ID MATCHES "(Clang|Gnu|GNU)")
   # TODO: Set for MSVC and other compilers.
   # TODO: Set -Werror when the build is clean.
 
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -std=gnu99")
```
dbaarda: We've actually been bitten by gnu extensions not working on some compilers/platforms in the past, so I'm kind of inclined to think we want c99, not gnu99. However, I'm not sure if that will break anything.

Tests suggest -std=c99 should be OK, but adding -pedantic spits out a bunch of warnings where we are not fully c99 compliant. We should probably fix that, but it gets a bit nasty because in some cases the way we are currently doing it will work for many Windows compilers that are not fully C99 compliant.

I think we probably want to make this -std=c99, and add -pedantic in another pull request that fixes all the non-C99 stuff, possibly adding conditional #define fixes for almost C99 compilers that we want to keep supporting.

texierp: Sure, thanks for the review @dbaarda !

You're indeed right, I will update this PR with '-std=c99'

dbaarda: Thanks for that, merging it now.

PR #182, opened 2019-11-26
Improve C99 compliance.

dbaarda:
Add  "-pedantic" to CMAKE_C_FLAGS for gcc and clang, and fix all the C99 warnings this produces.

In CMakeLists.txt remove all checks for includes and functions that are unused and/or are standard for C99. Add checks for `__func__` and `__FUNCTION__`. Remove explicit disable of non-standard HAVE_PROGRAM_INVOCATION_NAME. Update config.h.cmake to remove all cmakedefines that have never worked or have been removed, and re-order them to reflect the order in CMakeLists.txt.

Update checksum.h, version.c, and trace.[ch] to be C99 compliant. In trace.c remove support for compilers without varags macro support, and remove non-standard use of program_invocation_short_name. Note program_invocation_short_name was disabled in cmake anyway. Use C99 fputs() instead of Posix write(). Use RS_LOG_NONAME format when __func__ is an empty string. Use __func__ and __attribute__ selectively based on platform support.

Tidy up all #includes, removing unnecessary ones and re-arranging them into more consistent order. This includes no longer importing sys/types.h in librsync.h

Tidy indents/tabs in netint.c


dbaarda:
This prunes nearly 300 lines of old cruft. It should be fine on all the different platforms, but it would be good if anyone could test this on windows and/or BSD platforms to be sure.

dbaarda:
I also found/fixed that we've not been using 64bit support for calculating mdfour checksums! It seems at some point we stopped correctly setting HAVE_UINT64 which means we've been using only 32bit stuff for calculating md4sums. Doh! Ah well, we've mostly switched to blake2 now anyway.

PR #186, opened 2019-12-31
cmake/FindPOPT.cmake: Add PkgConfig support

ffontaine:
Add PkgConfig support to retrieve popt dependencies otherwise static
build can fail if popt has been linked with iconv:

```
[100%] Linking C executable rdiff
/home/buildroot/autobuild/instance-2/output-1/per-package/librsync/host/opt/ext-toolchain/bin/../lib/gcc/powerpc-buildroot-linux-uclibc/8.3.0/../../../../powerpc-buildroot-linux-uclibc/bin/ld: /home/buildroot/autobuild/instance-2/output-1/per-package/librsync/host/powerpc-buildroot-linux-uclibc/sysroot/usr/lib/libpopt.a(poptint.o): in function `POPT_fprintf':
poptint.c:(.text+0x34c): undefined reference to `libiconv_open
```

Fixes:
 - http://autobuild.buildroot.org/results/896e8e3efbedad90d66ae8c4e1e50f16206cab49

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/cmake/FindPOPT.cmake
+++ b/cmake/FindPOPT.cmake
@@ -42,20 +42,27 @@ if (NOT POPT_FOUND)
   endif (NOT POPT_ROOT_DIR)
 
   ##_____________________________________________________________________________
-  ## Check for the header files
+  ## Check with PkgConfig (to retrieve static dependencies such as iconv)
+  find_package(PkgConfig QUIET)
+  pkg_search_module (POPT QUIET popt)
+  if (NOT POPT_FOUND)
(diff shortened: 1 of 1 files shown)


dbaarda:
Thanks for this... it seems sad that every project needs their own version of this, and searching around it seems like they are all different and probably broken in different ways.

I fixed it a tiny bit more for the case where find_package() fails.

PR #187, opened 2020-01-09
fix various warnings

### LINKED ISSUE #184: Multiple "ISO C does not support the ‘I64’ ms_printf length modifier" warnings with MinGW-w64
With commit 0388468200ba18705c88e723893e19fabd5ba967, I get [multiple warnings](https://gist.github.com/Adsun701/bbfc449097a2e727df079948e6cba0f0) when building librsync with the MinGW-w64 runtime.

Adsun701:
This suppresses some harmless warnings given when compiling on MinGW-w64 due to differences in formatting. It also fixes some "dllimport" warnings and corrects a conditional block in `cmake/FindPOPT.cmake`. 

Fixes #184.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,14 +96,16 @@ check_type_size ( "unsigned long" SIZEOF_UNSIGNED_LONG )
 check_type_size ( "unsigned short" SIZEOF_UNSIGNED_SHORT )
 
 # Check for printf "%zu" size_t formatting support.
-if(CMAKE_CROSSCOMPILING)
+if(WIN32)
+  # CheckCSourceRuns checking for "%zu" succeeds but still gives warnings on win32.
+  set(HAVE_PRINTF_Z OFF)
+  # Not using unsupported %zu generates warnings about using %I64 with MinGW.
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-format")
+  message (STATUS "Compiling to Win32 - printf \"%zu\" size_t formatting support disabled")
+elseif(CMAKE_CROSSCOMPILING)
   # CheckCSourceRuns doesn't work when cross-compiling; assume C99 compliant support.
   set(HAVE_PRINTF_Z ON)
   message (STATUS "Cross compiling - assuming printf \"%zu\" size_t formatting support")
-elseif(WIN32)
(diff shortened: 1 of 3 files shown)


dbaarda:
So this fixes #184 by suppressing warnings about using "%I64" format strings on compilers that support the C99 "%uz" formatting we probably should be using.

We are using old windows "%I64" format strings in this case because when cross-compiling our CMakeLists.txt assumes we don't support "%zu" because we can't actively check for it. I have a feeling that it's probably better to assume that any compiler capable of cross-compiling is also C99 compliant and supports "%zu".

So I'm not sure about the "-Wno-format" part of this, but thanks heaps for the rest... I'm not sure how we missed correctly exporting the trace stuff.

dbaarda:
I've change the CMakeLists.txt to do what I think is a better solution for the format warnings. However, I don't have easy access to test that this actually works as I think it should.

Can you check if this is OK? If you are happy with this I'll merge it.

Adsun701:
There are [still warnings](https://gist.github.com/Adsun701/55734c086cc21926aecf2ea7ca8b2f5f) regarding "ISO C not supporting the I64 modifier". In addition, there are warnings regarding `unknown conversion type character ‘z’ in format`.

dbaarda:
Ugh, so it really doesn't support `"%uz"` but still complains about `"%I64"`, which we are also using for `FMT_LONG`. I guess it's using the target platform's implementation of printf.

I'll need to have a dig around to figure out the best solution. It would be nice if we could find a format specifier for these! Hat it doesn't complain about. Failing that we can silence the warnings for WIN32 only.

dbaarda:
The warnings it's giving about using %I64 for FMT_LONG comes from the #define here;

https://github.com/librsync/librsync/blob/24cef8f43f2708c6f4ddc68484ef07ee0a6f13bf/src/trace.h#L37

Which is basically using PRIdMAX defined in inttypes.h, which is part of the C99 standard. So it's warning about one of the C99 compatible headers provided for the WIN32 platform. It seems that using %I64 and %I is right for that platform, and the only thing we can do about the warnings for that platform is to turn them off.

I'm going to change this to turn of the warnings only for WIN32, and I'm going to simplify it to assume C99 compliance for other platforms.

dbaarda:
I finally got a chance to have another attempt at this. Can you please test it and make sure it works as intended?

Adsun701:
@dbaarda It works great!

PR #208, opened 2020-06-01
Add a Windows build target to Travis checks.

dbaarda:
This adds a windows build target to the Travis checks.

Unfortunately I couldn't find a MSVC compatible version of libpopt that could be easily installed in the Travis build image, so this cannot build and test rdiff. This may be an argument to migrate to a more widely used argument parser like GNU's gengetopt, which I believe is available for Windows on Travis via chocolatey in GnuWin.

This revealed some extra compatibility problems and compiler warnings from MSVC, some of which were genuine bugs that needed fixing. Many of these would have been found by gcc/clang's `-Wconversion` warnings, but they can be too noisy to turn on permanently for the old gcc used by Travis. The subset of `-Wsign-conversion` warnings are too spammy even for modern gcc/clang. I've made everything `-Wconversion -Wno-sign-conversion` compliant when using gcc 9.3.0 or clang 7.0.0, and enabled them in cmake for clang only.

Bugs/Fixes included are;

* Turn on `-Wconversion -Wno-sign-conversion` for clang and make all code compile clean.
* Configured cmake for MSVC compiler flags and made all code compile clean on Windows.
* Added cmake checking for `io.h` needed by MSVC for `fileutil.c`.
* Improved `fileutil.c` for MSVC, including `io.h` and preferring `_fileno()` to stop warnings.
* Fix broken error handling in `rs_file_copy_cb()` in `fileutil.c`.
* Improved trace output, making it less spamy and more consistent.
* Fix `patch.c` regression added in #206 that would break patching large files on 32bit architectures.
* Add checking deltas for invalid literal lengths of 0 or greater than SIZE_MAX.
* Tidy `stream.h`, removing unimplemented/unused functions and improving argument types.
* Delete `stream.c`, moving last `rs_buffers_copy()` to `rs_tube_copy_from_stream()` in `tube.c`.
* Remove not-required `LIBRSYNC_EXPORT` statements from `trace.c`.


dbaarda:
This is finally ready to merge. In the interest of preventing myself from fiddling with it any further I'm going to merge it now.

PR #212, opened 2020-08-30
Add gnuwin popt for travis windows tests.

dbaarda:
It appears the choco gnuwin should include popt, which should result in a working rdiff for testing on windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,18 @@ addons:
       - popt
     update: true
 
+cache:
+    directories:
+        - $HOME/AppData/Local/Temp/chocolatey
+
 script:
   - cmake --build . --target check
 
 matrix:
   include:
     - os: windows
+      before_install:
+        - choco install gnuwin
 
     - os: osx
       compiler: clang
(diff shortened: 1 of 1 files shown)


dbaarda:
Nope... that didn't work. It even spat out warnings that gnuwin is old and unmaintained, and you should probably use msys2 or cygwin.

Instructions on making travis use msys2 are here;

https://docs.travis-ci.com/user/reference/windows/#how-do-i-use-msys2

I also notice that #194 is happening again, and we are not seeing travis result ticks on github again. Sigh.

dbaarda:
The last attempt didn't work either, so I think the only way to make this work is to use the whole MSYS2 setup as documented here;

https://docs.travis-ci.com/user/reference/windows/#how-do-i-use-msys2

Alternatively we abandon popt in favor of a cmdline parser that's better supported, like getopt via gengetopt.

dbaarda:
This was a failed experiment, so I'm abandoning it. I think the best option is to switch from popt to a gengetopt generated getopt solution.

PR #213, opened 2020-09-21
Fix heap corruption from too small kbloom

ljusten:
kbloom is a bitmask with 'size2' bits, where 'size2' is the next power of 2 of 'size'. Thus, if 'size' is smaller than 4, 'size2' is smaller than 8, so that size2 / 8 == 0 and calloc allocates 0 bytes. This causes heap corruption when kbloom is subsequently written to. See discussion on https://groups.google.com/g/librsync/c/vmqzQS1QjIw.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -52,7 +52,7 @@ hashtable_t *_hashtable_new(int size)
     t->count = 0;
     t->tmask = size2 - 1;
 #ifndef HASHTABLE_NBLOOM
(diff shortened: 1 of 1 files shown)


dbaarda:
Note that the Bloom filter only helps if the hashtable is larger than L1 cache, which is about 8K entries. We could also just not allocate a bloom filter at all if the hash table is small and change the bloom filter setting/checking to do nothing if t->kbloom is NULL.

However, this is a much simpler fix for now, so I'm going to merge it.

PR #226, opened 2021-08-10
Tidy the cmake config and add better build configuration support.

dbaarda:
This is an idea taken from the following post;
    
https://blog.kitware.com/cmake-and-the-default-build-type/
    
It sets the default build type to Debug for git clone, otherwise the default is Release. It also sets possible values for the cmake-gui or ccmake. It works with multi-config generators and respects values set on the cmdline.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,10 @@ if (NOT CMAKE_SYSTEM_NAME)
 	message(FATAL_ERROR "No target OS set")
 endif()
 
+# Set CMAKE_BUILD_TYPE if unset.
+include(BuildType)
+message (STATUS "CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE}")
+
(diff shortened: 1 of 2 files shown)


dbaarda:
I have some mild concerns about this related to the following comment;

https://blog.kitware.com/cmake-and-the-default-build-type/#comment-8726

I don't know who might be affected by this though. Surely distro's relying on building using the default empty build-type have a workaround for projects using this approach.

I feel like the benefits of using this outweigh the risks. It means most naive users doing their own builds will end up with a Debug build with trace turned on, improving the chance they will detect problems. The improved cmake-gui and ccmake build-type selection will also make it easier for them to select a Release build when they realize they've been using a Debug build. Outputting the build-type in a cmake message makes it easier for them to realize/verify what build-type they are using.

Advanced users will already be setting CMAKE_BUILD_TYPE on the cmake cmdline, and this will not affect them.

So I think I'm going to submit this and we'll address any problems later if they surface.

PR #229, opened 2021-08-18
Add github action and make targets for clang-tidy and iwyu.

### LINKED ISSUE #215: Migrate from Travis to GitHub Actions.

dbaarda:
This adds clang-tidy and iwyu make targets for linting code, and adds a lint.yml githup action to run them.

Fixes #215 and replaces #222. This takes ideas from rizsotto's #222 pull request and changes them in the following ways;

1. it drops the `analyze-build` checks which research indicates are a subset of `clang-tidy`.
2. it sets `CMAKE_EXPORT_COMPILE_COMMANDS ON` in `CMakeLists.txt` so it doesn't need to be set at the cmdline.
3. it adds `clang-tidy` and `iwyu` make targets, so that they can easily be executed at the cmdline too.
4. it puts all the lint checks into a single github action, with each check as a separate step. This should be more efficient, and just as easy or easier to check the results.
5. Adds a `iwyu-fix` make target for automatically fixing headers using `fix_include`. This skips `fileutil.c` which confuses these tools by conditionally including all the available file related headers to try and find working functions for different platforms.
6. Fixes all warnings found by `clang-tidy`. There was actually not many.
7. Fixes all the include warnings found by `iwyu` except for `fileutil.c`. This included making the include guards consistently named, adding include guards to all the headers that didn't have them, and fixing all the includes.
8. Updates `CONTRIBUTING.md` to add a **Requirements** section listing dependencies and support tools to install, and how to use the new make targets.
9. Update `NEWS.md` with changes added so far.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,2 @@
+# clang-analyzer-valist.Uninitialized has a bug https://bugs.llvm.org/show_bug.cgi?id=41311.
+Checks: "-clang-analyzer-valist.Uninitialized"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,7 +13,7 @@ jobs:
       # Unfortunately available Marketplace Actions for this are in a mess, so we do it manually.
       run: |
         sudo apt-get update -y
-        sudo apt-get install -y libpopt-dev libb2-dev clang-tools iwyu
+        sudo apt-get install -y libpopt-dev libb2-dev clang-tidy iwyu
 
     - name: Configure CMake
       run: cmake .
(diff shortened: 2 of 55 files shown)


dbaarda:
I'm happy with this now, and I think I've dotted all the i's and crossed all the t's.

I'm going to merge this now.

PR #231, opened 2021-09-07
Upload build and install artifacts from Github actions

sourcefrog:
This makes downloadable artifacts out of the whole build tree and install results for every Check run. The build tree is perhaps useful for understanding build failures. The install results gives us compiled binaries and libraries that can be directly installed, which is particularly useful for platforms like Windows.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -40,8 +40,8 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/upload-artifact@v2
       with:
-        name: rdiff ${{ matrix.os }} ${{ matrix.build }}
-        path: build/CMakeFiles/rdiff*
+        name: build results ${{ matrix.os }} ${{ matrix.build }}
+        path: ${{github.workspace}}/build
 
     - name: Install Dependencies
       # Unfortunately available Marketplace Actions for this are in a mess, so we do it manually.
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -86,5 +86,5 @@ jobs:
         path: |
             ${{github.workspace}}/build/rdiff
             ${{github.workspace}}/build/rdiff.exe
-        if-no-files-found: error
-
+        // TODO: Skip this (to avoid warnings) in cases where rdiff is not expected to
+        // be built.
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -38,6 +38,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/upload-artifact@v2
+      with:
+        name: rdiff ${{ matrix.os }} ${{ matrix.build }}
+        path: build/rdiff*
 
     - name: Install Dependencies
       # Unfortunately available Marketplace Actions for this are in a mess, so we do it manually.
@@ -71,3 +75,4 @@ jobs:
     - name: Build docs
       if: ${{matrix.build_docs}}
       run: cmake --build '${{github.workspace}}/build' --target doc
+
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,7 +41,7 @@ jobs:
     - uses: actions/upload-artifact@v2
       with:
         name: rdiff ${{ matrix.os }} ${{ matrix.build }}
-        path: build/rdiff*
+        path: build/CMakeFiles/rdiff*
 
     - name: Install Dependencies
       # Unfortunately available Marketplace Actions for this are in a mess, so we do it manually.
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -82,7 +82,7 @@ jobs:
 
     - name: Upload rdiff
       uses: actions/upload-artifact@v2
-      if: ! ${{ matrix.no_rdiff }}
+      if: ${{ ! matrix.no_rdiff }}
       with:
         name: rdiff ${{ matrix.os }} ${{ matrix.build }}
         path: |
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,7 +19,6 @@ jobs:
        os: [ubuntu-latest, macos-latest, windows-latest]
        build: [Release]
        options: [""]
-       has_rdiff: true
        include:
          - os: windows-latest
            os_options: "-DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake"
@@ -31,7 +30,7 @@ jobs:
          - os: ubuntu-latest
            build: Release
            options: "-DBUILD_RDIFF=OFF"
-           has_rdiff: false
+           no_rdiff: false
          - os: ubuntu-latest
            build: Release
            options: "-G Ninja -DCMAKE_C_COMPILER=clang"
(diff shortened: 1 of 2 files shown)


.github/workflows/check.yml
```diff
@@ -71,3 +72,20 @@ jobs:
...
+        path: ${{github.workspace}}/build
+        if-no-files-found: error
+
+    - name: Upload rdiff
+      uses: actions/upload-artifact@v2
+      if: ${{ ! matrix.no_rdiff }}
+      with:
+        name: rdiff ${{ matrix.os }} ${{ matrix.build }}
+        path: |
+            ${{github.workspace}}/build/rdiff
```
dbaarda, line 89: Note that unless its a static build, the rdiff executable will also need the librsync library to run, which is either the .so (and related symlinks) on linux or the .dll on windows.

But perhaps, rather than manually selecting the bits to put in the artifact, perhaps we should do a `cmake --build '${{github.workspace}}/build' --target install` and then build an artifact out of everything that is installed. That way we should get the executables, supporting libraries, required headers to use those libraries, and supporting man pages.

That also means we don't need to selectively exclude the no-rdiff case, as we probably also want to build/check the artifact you get from installing only the library without rdiff.

I'm not sure what happens on windows when you run a 'cmake --target install`, and it wouldn't surprise me if it's broken. However, this is probably just something else we need to fix. 

The bad news is my experience with trying to make windows work for the github actions was kinda painful, but we are pretty close now to a working windows setup. At least github actions runs nice and fast for testing all this.

.github/workflows/check.yml
```diff
@@ -71,3 +72,20 @@ jobs:
...
+      with:
+        name: build results ${{ matrix.os }} ${{ matrix.build }}
+        path: ${{github.workspace}}/build
+        if-no-files-found: error
+
+    - name: Upload rdiff
+      uses: actions/upload-artifact@v2
+      if: ${{ ! matrix.no_rdiff }}
+      with:
+        name: rdiff ${{ matrix.os }} ${{ matrix.build }}
```
dbaarda, line 87: This is not enough to uniquely identify the particular job, as there is also an "options" dimension. This dimension means there are 4 different "os=ubuntu_latest,build=Release" runs with different options settings of "", "-DBUILD_SHARED_LIBS=OFF", "-DBUILD_RDIFF=OFF", and "-G Ninja -DCMAKE_C_COMPILER=clang".

I think this means it is currently building this artifact 4 times with each run and its a case of "last one is kept".

We probably want to keep all of them?

.github/workflows/check.yml
```diff
@@ -71,3 +72,20 @@ jobs:
...
       run: cmake --build '${{github.workspace}}/build' --target doc
+
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: build results ${{ matrix.os }} ${{ matrix.build }}
+        path: ${{github.workspace}}/build
+        if-no-files-found: error
+
+    - name: Upload rdiff
```
dbaarda, line 83: Perhaps call this "Upload install"?

.github/workflows/check.yml
```diff
@@ -71,3 +72,20 @@ jobs:
     - name: Build docs
       if: ${{matrix.build_docs}}
       run: cmake --build '${{github.workspace}}/build' --target doc
+
+    - name: Upload artifacts
```
dbaarda, line 76: Maybe call this "Upload build"?

.github/workflows/check.yml
```diff
@@ -71,3 +72,20 @@ jobs:
     - name: Build docs
       if: ${{matrix.build_docs}}
       run: cmake --build '${{github.workspace}}/build' --target doc
+
```
dbaarda, line 75: We might need a "Build install" step to build stuff for the install artifact... see below.

dbaarda (changes requested):
I have a few thoughts on this...

dbaarda:
The artifacts thing is interesting... I'm finding out more and more about what can be done with github actions and it's pretty cool. I'm starting to think about maybe having actions that automate much of the release process too.

One thing that worries me a little bit is... this is all a bit TOO cool... building/keeping all these artefacts for every checkin is surely going to eat resources. What are the freebie limits on all this? Is there some kind of retention-horizon or quota limit on these?

It would be nice to extend the release instructions in CONTRIBUTING.md to include uploading the windows binary as another "asset", the same way we grab a snapshot of the release tarball. People on Windows are far more likely to need a precompiled binary. I can update the CONTRIBUTING.md after we get this working nicely.

sourcefrog:
> One thing that worries me a little bit is... this is all a bit TOO cool... building/keeping all these artefacts for every checkin is surely going to eat resources. What are the freebie limits on all this? Is there some kind of retention-horizon or quota limit on these?

They're kept for 90 days by default: https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy

.github/workflows/check.yml
sourcefrog: I wonder if just building a self-contained static executable would be more helpful for people who want the executable?

sourcefrog:
> The artifacts thing is interesting... I'm finding out more and more about what can be done with github actions and it's pretty cool. I'm starting to think about maybe having actions that automate much of the release process too.
> 
> One thing that worries me a little bit is... this is all a bit TOO cool... building/keeping all these artefacts for every checkin is surely going to eat resources. What are the freebie limits on all this? Is there some kind of retention-horizon or quota limit on these?
> 
> It would be nice to extend the release instructions in CONTRIBUTING.md to include uploading the windows binary as another "asset", the same way we grab a snapshot of the release tarball. People on Windows are far more likely to need a precompiled binary. I can update the CONTRIBUTING.md after we get this working nicely.

Automating the release process sounds like a great idea, not only to reduce toil but also to move further towards https://security.googleblog.com/2021/06/introducing-slsa-end-to-end-framework.html

dbaarda:
I changed this to include the "build install" step and made the rdiff artifact into an "install" artifact. I also added the "options" dimension to the artifact name, so we can see all the build and install results.

I'm happy with this now... do you want to check it and see? 

How do you want to handle the merge? Normal merge, squash merge, rebase merge, or some kind of manual cleanup of the history? I'm not that fussed either way.

dbaarda (approved):
I think this is ready to merge... happy to let you do this in whatever way you think is best.

.github/workflows/check.yml
dbaarda: I changed this.

.github/workflows/check.yml
dbaarda: I changed this.

.github/workflows/check.yml
dbaarda: I added the matrix.options dimension to the name.

.github/workflows/check.yml
dbaarda: We now have install artifacts for all builds that we test. Currently, we only test static linking on Linux, so we don't have a static executable for windows, only a dynamic one. We could add another static build on windows, but I don't know if we need one.

dbaarda:
If I don't hear otherwise, I'm going to do a "Squash and merge" of this PR tomorrow.

dbaarda:
I'm going to squash and merge this now.

sourcefrog:
Thanks!

PR #234, opened 2021-09-14
Make delta directly process the input stream if it has enough data.

dbaarda:
This means deltas only accumulate data into the scoop buffer if the input stream is too small, otherwise it will process directly from the input stream. If the input stream always has enough input data to progress when calling rs_job_iter(), it will leave an unprocessed "tail" fragment in the input buffer that the caller must shuffle and refill. If there is not enough input data, it will start accumulating the data into the internal scoop buffer instead.

This makes delta calculations 5%~15% faster by avoiding all the data copying into the scoop buffer.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,19 @@
 
 NOT RELEASED YET
 
+ * Make delta directly process the input stream if it has enough data. Delta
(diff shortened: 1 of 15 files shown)


dbaarda:
I think this is good enough to merge... so I'm going to merge it now.

PR #243, opened 2023-01-31
Update github actions and fix `iwyu` build target.

dbaarda:
The old v2 versions of `checkout` and `upload-artifact` github actions now produce errors warning they must be updated. The new versions are v3.

The `iwyu` package has now fixed it's deps and we no longer need to explicitly install `libclang-common-9-dev` which doesn't exist any more.

The `iwyu` build target should ignore `fileutil.c`, and use neater clang output with noisy "note:" output removed.

Run `make iwyu-fix` to fix includes for `tests/rabinkarp_perf.c`.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,7 +37,7 @@ jobs:
            build: Debug
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Install Dependencies
       # Unfortunately available Marketplace Actions for this are in a mess, so we do it manually.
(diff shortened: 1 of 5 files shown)


dbaarda:
Note the lint check is now working correctly, but it's still failing because the iwyu test thinks the includes for fileutil.c and rabinkarp_perf.c are excessive.

In the case of fileutil.c, this is a platform-compatibility package that might not need some headers on linux, but might on other platforms. So this is a failure that can/should be ignored. Maybe we should somehow setup the tests to ignore this?

For rabinkarp_perf.c it's probably right, and it should be adjusted.

sourcefrog:
It looks like iwyu is failing in CI (https://github.com/librsync/librsync/actions/runs/14822488789/job/41611359535) in a way that I can not yet reproduce locally...

